PR #8, opened 2022-11-11
added benchmark related modules

shreyasXplain:


aixtend/modules/metric.py
```diff
@@ -0,0 +1,91 @@
+__author__='shreyassharma'
```
krishnadurai, line 1: File naming can be consistent: datasets.py, metric -> metrics.py, model -> models.py, and so forth.

aixtend/modules/model.py
```diff
@@ -189,4 +194,84 @@ def run_async(self, data: str, name: str = "model_process"):
...
+        Args:
+            model_info_json (dict): Json from API
+
+        Returns:
+            ModelInfo: Coverted ModelInfo object
+        """
+        m_info = ModelInfo(model_info_json['name'], model_info_json['id'], model_info_json['supplier']['id'])
+        return m_info
+
+    def get_models_from_page(self, page_number: int, task: str, input_language: str = None, output_language: str = None) -> List:
```
krishnadurai, line 211: I would prefer if we make a ModelFilter configuration and pass that to this class instead as we progress in the future. Don't address this now.

aixtend/modules/model.py
```diff
@@ -27,6 +27,11 @@
 import logging
 import traceback
 from requests.adapters import HTTPAdapter, Retry
+from collections import namedtuple
+from typing import List
+
+ModelInfo = namedtuple('ModelInfo', ['name', 'id', 'supplier'])
```
krishnadurai, line 33: Did you consider making all these info objects as classes? What if we have more fields that we need to accommodate in the future? Would a tuple be forwards compatible with more values?

aixtend/modules/benchmark.py
```diff
@@ -0,0 +1,154 @@
...
+                "samplesCount": 10
+                })
+            session = requests.Session()
+            retries = Retry(total=5,
+                            backoff_factor=0.1,
+                            status_forcelist=[ 500, 502, 503, 504 ])
+            session.mount('https://', HTTPAdapter(max_retries=retries))
+            r = session.post(url, headers=headers, data=payload)
+            resp = r.json()
+            logging.info(f"Creating Benchmark Job: Status of polling for {name} ({self.api_key}): {resp}")
```
krishnadurai, line 79: Change 'Status of polling' to 'Status'

aixtend/modules/benchmark.py
```diff
@@ -0,0 +1,154 @@
...
+            }
+            session = requests.Session()
+            retries = Retry(total=5,
+                            backoff_factor=0.1,
+                            status_forcelist=[ 500, 502, 503, 504 ])
+            session.mount('https://', HTTPAdapter(max_retries=retries))
+            r = session.get(url, headers=headers)
+            resp = r.json()
+            logging.info(f"Downloading Benchmark Results: Status of downloading results for {job_id} ({self.api_key}): {resp}")
+            csv_url = resp['reportUrl']
```
krishnadurai, line 143: Did we handle the case when the report isn't ready or present?

aixtend/modules/benchmark.py
```diff
@@ -0,0 +1,154 @@
...
+import time
+import json
+import requests
+import logging
+import traceback
+import pandas as pd
+from requests.adapters import HTTPAdapter, Retry
+from aixtend.utils.file_handler import download_file
+
+class Benchmark:
```
krishnadurai, line 32: Even though we have functions to create, start and download a benchmark, do we have a method to track a benchmark? Can we follow the same pattern that we in Model and Pipelines?
Here's a functional test depicting the same: https://github.com/aixplain/aiXtend/blob/main/tests/model_test.py

krishnadurai (commented):
@shreyasXplain good work. Here are a few comments apart from the ones that I have left in the review:
1. Can we add a functional test for running a benchmark?
2. Is there user-level documentation to run a benchmark? What is the intended user-experience for running a Benchmark job? It isn't quite clear to me how a user will make use of the [Model/Dataset/Metric]Info objects to run a Benchmark.

aixtend/utils/config.py
```diff
@@ -18,7 +18,9 @@
 
 PIPELINES_RUN_URL = os.getenv("PIPELINES_RUN_URL", "https://platform-api.aixplain.com/assets/pipeline/execution/run")
 MODELS_RUN_URL = os.getenv("MODELS_RUN_URL", "https://models.aixplain.com/api/v1/execute")
+BENCHMARKS_BACKEND_URL = os.getenv("BENCHMARKS_BACKEND_URL", "https://platform-api.aixplain.com")
 # GET THE API KEY FROM CMD
+TEAM_API_KEY = os.getenv("TEAM_API_KEY", "")
```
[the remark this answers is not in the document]
thiago-aixplain: @anasbakro, is it possible to unify the keys?

aixtend/utils/config.py
```diff
@@ -18,7 +18,9 @@
 
 PIPELINES_RUN_URL = os.getenv("PIPELINES_RUN_URL", "https://platform-api.aixplain.com/assets/pipeline/execution/run")
 MODELS_RUN_URL = os.getenv("MODELS_RUN_URL", "https://models.aixplain.com/api/v1/execute")
+BENCHMARKS_BACKEND_URL = os.getenv("BENCHMARKS_BACKEND_URL", "https://platform-api.aixplain.com")
```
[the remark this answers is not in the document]
thiago-aixplain: We set as a parameter to the class in case we want to specify domains (dev, test, etc) other than the production one, which is default.

aixtend/modules/model.py
thiago-aixplain: yes, this should be modeled as attributes of the class "Model" instead

aixtend/utils/config.py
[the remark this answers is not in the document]
krishnadurai: My suggestion is to hide this inside the class rather than having it as a parameter. The user can modify this through their environment.

aixtend/utils/config.py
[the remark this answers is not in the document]
krishnadurai: @anasbakro can we change the TEAM_API_KEY to have access to all models and pipelines that the user has access to?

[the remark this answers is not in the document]
shreyasXplain: @krishnadurai @anasbakro we would also need this to be available on the platform so that the user can get their keys

aixtend/modules/datasets.py
```diff
@@ -0,0 +1,116 @@
...
+            input_language (str, optional): Input language of listed datasets. Defaults to None.
+            output_language (str, optional): Output language of listed datasets. Defaults to None.
+
+        Returns:
+            List: List of datasets based on given filters
+        """
+        try:
+            dataset_info_list = []
+            assert k > 0
+            for page_number in range(k//10 + 1):
```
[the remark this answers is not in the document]
shreyasXplain: Yes, there are always 10 assets(datasets, models, metrics) per page, except on the last page where there can be less

aixtend/modules/model.py
```diff
@@ -189,4 +194,84 @@ def run_async(self, data: str, name: str = "model_process"):
...
+                if task in task_param_mapping["input"]:
+                    filter_params.append({"code" : task_param_mapping["input"][task], "value" : input_language})
+            if output_language is not None:
+                if task in task_param_mapping["ouput"]:
+                    filter_params.append({"code" : task_param_mapping["ouput"][task], "value" : output_language})
+            headers = {
+            'Authorization': f"Token {self.api_key}",
+            'Content-Type': 'application/json'
+            }
+            session = requests.Session()
```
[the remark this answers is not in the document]
shreyasXplain: Yes, I can add that

aixtend/modules/model.py
shreyasXplain: We will add the information to the module as attributes and shift the functional uses(listing etc) inside factories as suggested by @thiago-aixplain

krishnadurai:
@shreyasXplain can you please update the PR?

shreyasXplain:
> @shreyasXplain can you please update the PR?

@krishnadurai I am currently not working on the SDK. We might pick it up next week(depending on the priorities)

aixtend/utils/file_handler.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Args:
+        download_url (str): URL of file to download
+        download_file_path (str, optional): File path to save downloaded file. If None then generates random path is tempdirectory. Defaults to None.
+
+    Returns:
+        str: Path where file was downloaded
+    """
+    if download_file_path is None:
+        sys_temp_dir = tempfile.gettempdir()
```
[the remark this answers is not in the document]
shreyasXplain: Is there a way to have the temp file persist even after code execution finishes? A user will probably want to access the file later as well. The download path will now be something like '/tmp/aiXtend/filename.ext'. Do you have suggestions on if we should save it someplace else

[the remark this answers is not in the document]
krishnadurai: Tempdir and tempfile are both the same in terms of retaining.
Why do you see that a download utility function needs the file to be retained after it is processed? A user can download it again if they want to check the contents. Whereas a library has to be cautious about disk usage and cache items.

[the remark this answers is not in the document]
shreyasXplain: tempfile.gettempdir() just gives the default temp folder of the machine and doesn't actually create a temporary directory or file.
 This is for the functionality of downloading the results of the benchmarks as a CSV so it needs to be persistent. In case a save path isn't provided we use this path.

aixtend/factories/metric_factory.py
```diff
@@ -0,0 +1,76 @@
...
+            List[Metric]: List of supported metrics
+        """
+        try:
+            url = f"{self.backend_url}/sdk/scores?function={task}"
+            headers = {
+            'Authorization': f"Token {self.api_key}",
+            'Content-Type': 'application/json'
+            }
+            r = _request_with_retry("get", url, headers=headers)
+            resp = r.json()
```
anasbakro, line 68: My suggestion is that there should be a class called BE API which has a method that gets or post or put and takes a path as a parameter something like BE_API.post("scores?function={task}") and internally it will do the call and pick up the token and URL from the env

anasbakro (approved):

shreyasXplain:
@krishnadurai Could you also please take a look at the new changes. We have made major changes to the structure and design. Few small things to add before the final merge

- Adding tests with @lucas-aixplain 
- Making the factories methods as all static(?)
- Some User Doc to help them get started

aixtend/factories/metric_factory.py
krishnadurai: @shreyasXplain this comment is unresolved.

shreyasXplain: @anasbakro This would be okay for the team-api key but for model api-key we would still need to pass the API key(for model/pipeline) as a parameter.  Also we would need to still have this function to send normal requests for tasks like downloading files etc.
Nonetheless we can still in addition to current function develop the class. What do you think?

krishnadurai:
Please resolve the conflicts as well

tests/asset_factory_test.py
```diff
@@ -0,0 +1,127 @@
...
+    task = "test_task"
+    url_dict = {
+        "model" : f"{config.BENCHMARKS_BACKEND_URL}/sdk/inventory/models/?pageNumber={page_number}&function={task}",
+        "dataset": f"{config.BENCHMARKS_BACKEND_URL}/sdk/datasets?pageNumber={page_number}&function={task}"
+    }
+    for asset_name in asset_list:
+        AssetFactory = __get_asset_factory(asset_name)
+        with requests_mock.Mocker() as mock:
+            url = url_dict[asset_name]
+            mock_json = json.load(open(Path("tests\mock_responses\list_assets_responses.json")))[asset_name]
```
krishnadurai, line 57: Always use `with open`. Please edit all your tests to reflect the same.

tests/asset_factory_test.py
```diff
@@ -0,0 +1,127 @@
...
+        AssetFactory = __get_asset_factory(asset_name)
+        with requests_mock.Mocker() as mock:
+            url = url_dict[asset_name]
+            mock_json = json.load(open(Path("tests\mock_responses\get_asset_info_responses.json")))[asset_name]
+            mock.get( url, headers=FIXED_HEADER, json=mock_json)
+            asset = AssetFactory.create_asset_from_id(asset_id)
+        assert asset.get_asset_info()['id'] == asset_id
+
+def test_asset_listing_from_page():
+    asset_list = ["model", "dataset"]
```
krishnadurai, line 45: Is 'metric' intentionally ignored?

tests/pipeline_test.py
```diff
@@ -1,4 +1,4 @@
-__author__='thiagocastroferreira'
+__author__ = "thiagocastroferreira"
```
krishnadurai, line 1: Can all assets: metrics, benchmarks, models have a functional test similar to this?

The code will be similar to your unit tests, though it will test the e2e functionality of the module. This can be made a part of a later PR.

tests/benchmark_test.py
```diff
@@ -0,0 +1,125 @@
...
+        benchmarkJob = BenchmarkFactory.create_benchmark_job_from_id(asset_id)
+        download_url = json.load(open(Path("tests\mock_responses\get_asset_info_responses.json")))['benchmarkJob']['reportUrl']
+        dummy_result_path = Path("tests\mock_responses\dummy_result.csv")
+        dummy_content = open(dummy_result_path, "rb").read()
+        mock.get(download_url, headers=FIXED_HEADER, content=dummy_content)
+        result_path_without_save_path = BenchmarkFactory.download_results_as_csv(benchmarkJob)
+        result_path_with_save_path = BenchmarkFactory.download_results_as_csv(benchmarkJob, "test_save_results.csv")
+        result_df_without_save_path = BenchmarkFactory.download_results_as_csv(benchmarkJob, returnDataFrame=True)
+    
+    result_path_without_save_path_content = open(result_path_without_save_path, 'rb').read()
```
krishnadurai, line 112: with open

krishnadurai (commented):
This is great @shreyasXplain, just 2 or 3 quick fixes are suggested in this PR.
Did you update the setup file with the new requirements?

krishnadurai:
@shreyasXplain is this PR ready?

shreyasXplain:
@krishnadurai Yes, the PR is ready. We will add and reorg the test in a new PR

krishnadurai (approved):

PR #16, opened 2023-03-09
Enabling parameters input options in models

thiago-aixplain:


aixtend/modules/model.py
```diff
@@ -186,15 +190,17 @@ def run(self, data: Union[str, dict], name: str = "model_process", timeout: floa
             end = time.time()
             return {"status": "FAILED", "error": msg, "elapsed_time": end - start}
 
-    def run_async(self, data: Union[str, dict], name: str = "model_process") -> dict:
+
+    def run_async(self, data: Union[Text, Dict], parameters: Dict = {}, name: Text = "model_process") -> Dict:
```
krishnadurai, line 194: Minor change: It would be always better to add new function parameters to the end of the list of arguments to be sent to ensure backward compatibility. Keep this in mind for the future.

thiago-aixplain: Thank you @krishnadurai , I just address the problem.

krishnadurai (approved):

krishnadurai (approved):

PR #28, opened 2023-04-25
Dataset Onboard and List Corpus/Dataset Services

thiago-aixplain:


shreyasXplain (approved):
Looks good @thiago-aixplain! 
- We just need to discuss about all of us doing some testing and deploying to pypi - especially because the current deployed version links to main for documentation
- Also, might want to add some more explanation later, differentiating between corpus and dataset(might be difficult for a new person coming in and looking at it for the first time)

PR #31, opened 2023-05-11
Update streaming client example to stream bytes directly from file instead of using pydub

davidpodc:


docs/streaming/README.md
```diff
@@ -40,3 +40,24 @@ python3.8 aixplain_diarization_streaming_client.py --file-path=test_dia.wav --ca
...
+
+Our service is configured to process audio streamed as a single channel with a sampling rate of 16000Hz.
+
+If the audio does not meet these specifications, the service may yield unexpected results.
+
+To ensure compatibility, we provide a helper script that adjusts your audio files to meet these requirements.
+
+### Installing Dependencies
+
+The helper script relies on the pydub package. Install it using pip:
```
krishnadurai, line 55: Can we make it clearer that only the compatibility conversion relies on it?

krishnadurai (approved):

PR #40, opened 2023-06-01
M 4551591243 image dataset

thiago-aixplain:


aixplain/factories/corpus_factory.py
```diff
@@ -90,6 +90,11 @@ def __from_response(cls, response: Dict) -> Corpus:
         except:
             license = None
 
+        try:
+            length = int(response["segmentsCount"])
+        except:
```
krishnadurai, line 95: Correct to: `except Exception:`
Is segment count a necessary parameter to be returned in the response? How about logging the contents of the response if segmentsCount is not retrieved?

aixplain/factories/dataset_factory.py
```diff
@@ -133,6 +133,11 @@ def __from_response(cls, response: Dict) -> Dataset:
         except:
             license = None
 
+        try:
+            length = int(response["segmentsCount"])
+        except:
```
krishnadurai, line 138: Same here. Log here too.

aixplain/processes/data_onboarding/process_media_files.py
```diff
@@ -0,0 +1,223 @@
...
+    Args:
+        metadata (MetaData): meta data of the asset
+        paths (List): list of paths to local files
+        folder (Path): local folder to save compressed files before upload them to s3.
+
+    Returns:
+        Tuple[List[File], int, int, int]: list of s3 links; data, start and end columns index, and number of rows
+    """
+    # if files are stored locally, create a folder to store it
+    media_folder = Path(".")
```
krishnadurai, line 45: Usually, a good practice is to use a [temporary folder](https://docs.python.org/3/library/tempfile.html).
You can use a with context to create this folder and store the files in the tempdir. I would recommend that you delete this after uploading its contents.

aixplain/processes/data_onboarding/process_media_files.py
```diff
@@ -0,0 +1,223 @@
...
+        media_folder.mkdir(exist_ok=True)
+
+    idx = 0
+    data_column_idx, start_column_idx, end_column_idx = -1, -1, -1
+    files, batch, start_intervals, end_intervals = [], [], [], []
+    for i in tqdm(range(len(paths)), desc=f' Data "{metadata.name}" onboarding progress', position=1, leave=False):
+        path = paths[i]
+        try:
+            dataframe = pd.read_csv(path)
+        except Exception as e:
```
krishnadurai, line 57: Is it necessary to catch the exception and re-raise it?

aixplain/processes/data_onboarding/process_media_files.py
```diff
@@ -0,0 +1,223 @@
...
+            except Exception as e:
+                message = f'Data Asset Onboarding Error: Column "{metadata.name}" not found in the local file "{path}".'
+                logging.exception(message)
+                raise Exception(message)
+
+            # adding medias
+            if metadata.storage_type == StorageType.FILE:
+                if metadata.dtype == DataType.AUDIO:
+                    assert (
+                        os.path.getsize(media_path) <= 50000000
```
krishnadurai, line 76: Declare all constants right after the imports in the top of the file and name them.

aixplain/processes/data_onboarding/process_media_files.py
```diff
@@ -0,0 +1,223 @@
...
+                    ), f'Data Asset Onboarding Error: Local audio file "{media_path}" exceeds the size limit of 50 MB.'
+                else:
+                    assert (
+                        os.path.getsize(media_path) <= 25000000
```
krishnadurai, line 80: Same here.

aixplain/processes/data_onboarding/process_media_files.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def compress_folder(folder_path: str):
+    with tarfile.open(folder_path + ".tgz", "w:gz") as tar:
+        for name in os.listdir(folder_path):
+            tar.add(os.path.join(folder_path, name))
+    return folder_path + ".tgz"
+
+
+def run(metadata: MetaData, paths: List, folder: Path, batch_size: int = 100) -> Tuple[List[File], int, int, int, int]:
```
krishnadurai, line 27: Can these definitions be combined?
https://github.com/aixplain/aiXplain/blob/main/aixplain/processes/data_onboarding/process_audio_files.py#L26

krishnadurai (commented):
I've left behind some comments.
@thiago-aixplain I think the compress and upload definition can be combined into a single upload functionality. Can you please give it a thought?

krishnadurai (approved):

PR #45, opened 2023-06-19
 executing tests for each push

MAlyafeai18:
* adding GitHub action to execute the tests of the GitHub.
* currently the passing of tests is not put into consideration I just made sure that the tests are working 
* Lastly we need to add secrets to GitHub repo with the name TEAM_API_KEY to a value to one of the APIs keys of the platform after that the Github actions can be executed normally

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -1,0 +1,28 @@
+name: Python Tests
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Run Pytest
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8]
+
+    steps:
+    - uses: actions/checkout@v2
+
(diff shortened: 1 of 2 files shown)


.github/workflows/python-test.yml
```diff
@@ -0,0 +1,31 @@
+name: Python Tests
+
+on: [push, pull_request]
```
krishnadurai, line 3: Can you also please enable the tests before merge?
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions

krishnadurai (commented):
@AlyaF3i are we also going to enable the functional tests marked as skip in the tests that we have?
Can you please generate a list of tests that we need to implement for E2E, functional, and unit tests?

krishnadurai:
@AlyaF3i is this PR going to be worked upon?

PR #61, opened 2023-08-13
validating input of dataset onboarding

MAlyafeai18:
the output can be done as well the issue is that the backend doesn't send the output datatype, so I want to suggest to modify the backend endpoint which send the functions and include the output datatype. we can discuss this further during the meeting

### DIFF
--- a/aixplain/factories/dataset_factory.py
+++ b/aixplain/factories/dataset_factory.py
@@ -36,7 +36,7 @@ from aixplain.modules.metadata import MetaData
 from aixplain.enums.data_subtype import DataSubtype
 from aixplain.enums.data_type import DataType
 from aixplain.enums.error_handler import ErrorHandler
-from aixplain.enums.function import Function
+from aixplain.enums.function import Function, FunctionInput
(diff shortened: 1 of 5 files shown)


aixplain/factories/dataset_factory.py
```diff
@@ -360,6 +360,11 @@ def create(
         Returns:
             Dict: dataset onboard status
         """
+        input_dtype = input_schema[0].dtype if isinstance(input_schema[0],MetaData) else input_schema[0]['dtype']
+        # output_dtype = output_schema[0].dtype if isinstance(output_schema[0],MetaData) else output_schema[0]['dtype']
+        assert (
+            FunctionInput.get(function) is not None and FunctionInput[function]['input'] == input_dtype
+        ), f"Input data type is not compatible with the Function {function}"
```
thiago-aixplain, line 367: Just rephrase the message:

```
Data Asset Onboarding Error: Input data type ({input_dtype}) is not compatible with the Function {function}
```

thiago-aixplain (approved):

PR #62, opened 2023-08-14
Function tests for pipelines

thiago-aixplain:


shreyasXplain (approved):

krishnadurai (commented):
Thanks @thiago-aixplain, can we please move these tests to the tests/functional/ folder?

PR #67, opened 2023-08-23
BugFix: update function filtering in model listing

shreyasXplain:


krishnadurai (commented):
Let’s add a test that catches this bug.

shreyasXplain:
> Let’s add a test that catches this bug.
@krishnadurai @thiago-aixplain 
We don't need have info like function or language as attributes of the metric class. We might want to add these info to make this validation possible.

aixplain/modules/model.py
```diff
@@ -76,6 +77,12 @@ def __init__(
         self.additional_info = additional_info
         self.url = config.MODELS_RUN_URL
         self.backend_url = config.BACKEND_URL
+        if "function" in additional_info:
```
thiago-aixplain, line 80: @shreyasXplain to convert a string into Enum, you may do it:

```
self.function = Function(additional_info["function"])
```

aixplain/modules/model.py
```diff
@@ -76,6 +77,12 @@ def __init__(
...
+            self.function = None
```
thiago-aixplain, line 81: Add function in the docstring of model asset:

https://github.com/aixplain/aiXplain/blob/M-5038717082-ModelFilterFix/aixplain/modules/model.py#L46

thiago-aixplain (changes requested):

aixplain/modules/model.py
shreyasXplain: But that needs function in all caps i.e. `Function["TRANSLATION"]` will work but `Function["translation"]` will not.

thiago-aixplain (approved):

PR #69, opened 2023-09-06
Fix the size limit to reflect 1 GB

thiago-aixplain:


aixplain/processes/data_onboarding/onboard_functions.py
```diff
@@ -68,7 +68,7 @@ def get_paths(input_paths: List[Union[str, Path]]) -> List[Path]:
     # check CSV sizes
     for path in paths:
         assert (
-            os.path.getsize(path) <= 100000000
```
lucas-aixplain, line 71: @thiago-aixplain, I think it's better to write `1e9` instead of multiple 0's. Also, you could create a constant variable instead of putting the value directly.

lucas-aixplain (approved):

PR #73, opened 2023-09-13
Just get data tyoe from input required items

thiago-aixplain:


MAlyafeai18 (changes requested):
looks good, however I believe it's better if we kept it as set instead of convert it to list, it will be better for memory and for searching as well.
where as for the part where you check if it's required this looks fine

aixplain/enums/function.py
```diff
@@ -50,7 +50,7 @@ def load_functions():
     functions_input_output = {
         function["id"]: {
             "input" : {
-                input_data_object["dataType"] for input_data_object in function["params"]
+                input_data_object["dataType"] for input_data_object in function["params"] if input_data_object["required"] is True
             },
             "output" : {
                 output_data_object["dataType"] for output_data_object in function["output"]
```
lucas-aixplain, line 56: @thiago-aixplain, why checking the `required` field is only done for input and not for output?

lucas-aixplain (approved):

MAlyafeai18 (approved):

PR #85, opened 2023-10-05
Removing problematic finetuning test

mikelam-us-aixplain:
Removing problematic fine-tuning test due to AWS rate-limiting until this problem is fixed.

### DIFF
--- a/tests/functional/finetune/finetune_functional_test.py
+++ b/tests/functional/finetune/finetune_functional_test.py
@@ -47,7 +47,7 @@ def run_input_map(request):
 def list_input_map(request):
     return request.param
 
-'''
+@pytest.mark.skip(reason="AWS rate limiting causing problems")
 def test_run(run_input_map):
     model = ModelFactory.get(run_input_map["model_id"])
     dataset_list = [DatasetFactory.get(run_input_map["dataset_id"])]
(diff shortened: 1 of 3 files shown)


tests/functional/finetune/finetune_functional_test.py
```diff
@@ -47,7 +47,7 @@ def run_input_map(request):
 def list_input_map(request):
     return request.param
 
-
+'''
```
krishnadurai, line 50: Can you mark this test as skipped?

lucas-aixplain: @mikelam-us-aixplain, to remove AWS ASR from the tests, you should erase AWS ASR entry from the `finetune_test_run_data.json` file. It is the second element: https://github.com/aixplain/aiXplain/blob/M-5283031693-Removing-Problematic-Finetune_Test/tests/functional/finetune/data/finetune_test_run_data.json#L7C22-L8
Please also save a copy of the file with all suppliers as a backup.

lucas-aixplain (changes requested):

tests/functional/finetune/finetune_functional_test.py
krishnadurai: @lucas-aixplain why remove when this can be marked as skipped? We can leave a note to why it was marked as skipped.

lucas-aixplain (approved):

lucas-aixplain (approved):

krishnadurai (approved):

PR #86, opened 2023-10-09
File Upload

MAlyafeai18:
Still Working on it, there is an issue with the autorization so I am waiting for a response from Hadi


aixplain/factories/file_factory.py
```diff
@@ -53,14 +60,25 @@ def upload(cls, local_path: Text) -> Text:
...
         else:
             ftype = mime_type.split("/")[0]
         if os.path.getsize(local_path) > type_to_max_size[ftype]:
-            raise Exception(f'File Upload Error: local file "{local_path}" of type "{mime_type}" exceeds {type_to_max_size[ftype] / MB_1} MB.')
-        s3_path = upload_data(file_name=local_path)
+            raise Exception(
+                f'File Upload Error: local file "{local_path}" of type "{mime_type}" exceeds {type_to_max_size[ftype] / MB_1} MB.'
+            )
+
+        if not is_temp:
```
thiago-aixplain, line 78: `if is_temp is False`

But why do you need this condition?

What is the difference between `os.path.basename(local_path)` and `local_path`

aixplain/factories/file_factory.py
```diff
@@ -104,3 +122,26 @@ def to_link(cls, data: Union[Text, Dict]) -> Union[Text, Dict]:
...
+            license (License, optional): the license for the file
+            is_temp (bool): specify if the file that will be upload is a temporary file
+
+        Returns:
+            Text: The S3 path where the file was uploaded.
+
+        Raises:
+            FileNotFoundError: If the local file is not found.
+            Exception: If the file size exceeds the maximum allowed size.
+        """
```
thiago-aixplain, line 145: Add an assertion that check whether license is not None if is_temp is not equals to False

aixplain/utils/file_utils.py
```diff
@@ -103,7 +113,12 @@ def upload_data(
     try:
         # Get pre-signed URL
         # URL of aiXplain service which returns a pre-signed URL to onboard the file
-        url = urljoin(config.BACKEND_URL, "sdk/file/upload/temp-url")
+        if is_temp:
```
thiago-aixplain, line 116: `if is_temp is True`

aixplain/factories/file_factory.py
```diff
@@ -53,14 +60,25 @@ def upload(cls, local_path: Text) -> Text:
...
             ftype = mime_type.split("/")[0]
         if os.path.getsize(local_path) > type_to_max_size[ftype]:
-            raise Exception(f'File Upload Error: local file "{local_path}" of type "{mime_type}" exceeds {type_to_max_size[ftype] / MB_1} MB.')
-        s3_path = upload_data(file_name=local_path)
+            raise Exception(
+                f'File Upload Error: local file "{local_path}" of type "{mime_type}" exceeds {type_to_max_size[ftype] / MB_1} MB.'
+            )
+
+        if is_temp is False:
+            s3_path = upload_data(file_name=os.path.basename(local_path), tags=tags, license=license, is_temp=is_temp, content_type=content_type)
```
thiago-aixplain, line 79: If the user gives a relative path (e.g. `folder/file.txt`), the document will not be found, since we are using just the base name (`file.txt`). So this is a bug and may be fixed.

thiago-aixplain (changes requested):
The unit tests are not passing after this change. Can you take a look?

Also please update the documentation here: https://github.com/aixplain/aiXplain/blob/main/docs/user/user_doc.md#assets

Create a new session called `Files`

thiago-aixplain:
The functional tests are not passing yet

thiago-aixplain (approved):

PR #103, opened 2023-12-21
Create bounds for FineTune hyperparameters

lucas-aixplain:
Task: M-5725704280

### DIFF
--- a/aixplain/modules/finetune/hyperparameters.py
+++ b/aixplain/modules/finetune/hyperparameters.py
@@ -13,19 +13,52 @@ class SchedulerType:
     REDUCE_ON_PLATEAU = "reduce_lr_on_plateau"
 
 
+EPOCHS = [1]
+BATCH_SIZE = [1]
+
+
 @dataclass_json
 @dataclass
 class Hyperparameters(object):
-    epochs: int = 4
-    train_batch_size: int = 4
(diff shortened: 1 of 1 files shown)


aixplain/modules/finetune/hyperparameters.py
```diff
@@ -13,19 +13,52 @@ class SchedulerType:
     REDUCE_ON_PLATEAU = "reduce_lr_on_plateau"
 
 
+EPOCHS = [1]
```
thiago-aixplain, line 16: Maybe another name like `EPOCH_VALUES`.

lucas-aixplain:
We're only allowing the users to utilize working hyperparameters. Apart from the ones that are [set by default in Flyte](https://github.com/aixplain/aaron-experiments/blob/master/flyte/finetune-llm/src/finetune/data.py#L8), we have [here](https://github.com/aixplain/finetuner/blob/development/language_model/config/fixed_hyperparameters.json) the fixed hyperparameters.

thiago-aixplain (changes requested):
The changes look fine and I approve them. However, now we do have `development` and `test` branches for the SDK. So please assign the PR with the `development` branch as the base one.

thiago-aixplain (approved):

PR #125, opened 2024-02-07
Update Finetuner hyperparameters

lucas-aixplain:
- Add `train_batch_size` and `eval_batch_size` as available parameters for fine-tuning LLMs.
- Add constraints to limit possible values based on conducted tests.
- Add unit test for hyperparameters.

Task: M-6021007588


aixplain/modules/finetune/hyperparameters.py
```diff
@@ -56,6 +65,12 @@ def __post_init__(self):
         if self.epochs > EPOCHS_MAX_VALUE:
             raise ValueError(f"epochs must be one less than {EPOCHS_MAX_VALUE}")
```
thiago-aixplain, line 66: less or equal

aixplain/modules/finetune/hyperparameters.py
```diff
@@ -56,6 +65,12 @@ def __post_init__(self):
...
 
+        if self.train_batch_size not in BATCH_SIZE_VALUES:
+            raise ValueError(f"train_batch_size must be one of the following values: {BATCH_SIZE_VALUES}")
+
+        if self.eval_batch_size not in BATCH_SIZE_VALUES:
+            raise ValueError(f"eval_batch_size must be one of the following values: {BATCH_SIZE_VALUES}")
+
         if self.max_seq_length > MAX_SEQ_LENGTH_MAX_VALUE:
```
thiago-aixplain, line 74: less or equal

thiago-aixplain (approved):

aixplain/modules/finetune/hyperparameters.py
lucas-aixplain: In this case, we should only raise ValueError if the number of epochs is greater than the maximum allowed value (`EPOCHS_MAX_VALUE`), so the logic is correct.

aixplain/modules/finetune/hyperparameters.py
lucas-aixplain: For batch size, I want only to receive values from the list BATCH_SIZE_VALUES, that's why we have `not in` there.

aixplain/modules/finetune/hyperparameters.py
lucas-aixplain: I'll change the message to ensure it's correct.

PR #132, opened 2024-02-15
Fix supplier representation of a model

thiago-aixplain:


aixplain/modules/model.py
```diff
@@ -95,7 +95,10 @@ def to_dict(self) -> Dict:
         return {"id": self.id, "name": self.name, "supplier": self.supplier, "additional_info": clean_additional_info}
 
     def __repr__(self):
-        return f"<Model: {self.name} by {self.supplier}>"
+        try:
+            return f"<Model: {self.name} by {self.supplier['name']}>"
```
lucas-aixplain, line 99: Can `self.supplier` be `Text` or `Dict`? If so, we should change the `Model` constructor documentation to add `Union` to `Dict`.

lucas-aixplain (approved):

PR #133, opened 2024-02-21
M 6107719447 check finetuner status

thiago-aixplain:


aixplain/modules/finetune/status.py
```diff
@@ -0,0 +1,36 @@
...
+from aixplain.enums.asset_status import AssetStatus
+from dataclasses import dataclass
+from dataclasses_json import dataclass_json
+from typing import Optional, Text
+
+@dataclass_json
+@dataclass
+class FinetuneStatus(object):
+    status: "AssetStatus"
+    model_status: "AssetStatus"
```
lucas-aixplain, line 33: Here it should be:
```python
status: AssetStatus
model_status: AssetStatus
```

aixplain/modules/model.py
```diff
@@ -251,23 +250,65 @@ def run_async(self, data: Union[Text, Dict], name: Text = "model_process", param
                 response["error"] = msg
         return response
 
-    def check_finetune_status(self):
+    def check_finetune_status(self, after_epoch: Optional[int] = None):
```
lucas-aixplain, line 253: Please add `FinetuneStatus` in the return annotation:
```python
def check_finetune_status(self, after_epoch: Optional[int] = None) -> FinetuneStatus:
```

lucas-aixplain (approved):

PR #145, opened 2024-03-14
M 5905660469 enhance benchmark job response

shreyasXplain:


aixplain/factories/benchmark_factory.py
```diff
@@ -237,3 +238,24 @@ def list_normalization_options(cls, metric: Metric, model: Model) -> List[str]:
...
             logging.error(error_message, exc_info=True)
             return []
+
+    @classmethod
+    def get_benchmark_job_scores(cls, job_id):
+        def __get_model_name(model_id):
+            model = ModelFactory.get(model_id)
+            supplier = str(model.supplier)
+            try:
+                if isinstance(supplier, Supplier):
```
thiago-aixplain, line 248: Supplier is a trick field. Check the different ways we handled in here to avoid bugs:

https://github.com/aixplain/aiXplain/blob/main/aixplain/modules/asset.py#L53C9-L61C14

aixplain/modules/benchmark_job.py
```diff
@@ -92,3 +92,70 @@ def download_results_as_csv(self, save_path: Optional[Text] = None, return_dataf
...
             raise Exception(error_message)
+        
+    def __simplify_scores(self, scores):
+        simplified_score_list  = []
+        for model_id, model_info in scores.items():
+            model_scores = model_info["rawScores"]
+            # model = Mode
+            row = {"Model": model_id}
+            for score_info in model_scores:
+                row[score_info["longName"]] = score_info["average"]
```
thiago-aixplain, line 103: Make sure these fields are always there. Does it make sense to add a try/catch to treat when they are not?

aixplain/factories/benchmark_factory.py
shreyasXplain: I saw a case where supplier was a dict but in form of a string, hence the extra condition. Since we do not work on raw data from BE API, all the cases handled in model factory make sure that supplier is of instance Supplier or a string in worst case. So benchmark factory does not need to again do the same.

aixplain/modules/benchmark_job.py
shreyasXplain: These are mandatory fields and will always be there. In case of any issue, they are sent as None instead but the key will always be there

thiago-aixplain (approved):

PR #157, opened 2024-04-01
Interval label new structures

thiago-aixplain:


shreyasXplain (commented):
@thiago-aixplain I'm assuming this is like x, y, w and h. Is that right?
If so, should we be more explicit in the naming - as this might be still clear. Also saying what is this with object detection examples instead of image classification - maybe one would be good

shreyasXplain (approved):

PR #169, opened 2024-04-18
SDK's UML diagram

thiago-aixplain:


docs/assets/architecture.png
lucas-aixplain, line 1: We need to update the diagram to reflect the new assets. From what I looked in the repo, it's missing Finetune, Corpus, and File modules and factories. Also, the title at the top should be "aiXplain" instead of "aiXtend".

lucas-aixplain (approved):

PR #208, opened 2024-07-24
ENG-1: programmatic api introduced

kadirpekel:
Includes a samples directory, unit tests and functional tests.

Only routing sample works but for text input only, audio input receives a backend error such below which needs to be clarified

```
{
    "completed": true,
    "error": "err.supplier_error",
    "supplierError": "IN_PROGRESS"
}
```

Routing example itself:

```
from aixplain.api import Pipeline

pipeline = Pipeline()

# add nodes to the pipeline
input = pipeline.input()
translation = pipeline.asset("60ddefae8d38c51c5885eff7")
speech_recognition = pipeline.asset("621cf3fa6442ef511d2830af")

# route the input to the translation and speech recognition nodes
input.route(translation.inputs.text, speech_recognition.inputs.source_audio)

# route the output of the translation and speech recognition nodes
# to the output
translation.use_output("data")
speech_recognition.use_output("data")

# save the pipeline as draft
pipeline.save()

# run the pipeline
output = pipeline.run("hello programmatic pipeline!")

# print the output
print(output)
```


aixplain/api/base.py
```diff
@@ -0,0 +1,317 @@
...
+            self.to_node = self.to_node.number
+
+    def attach(self, pipeline: "Pipeline"):
+        """
+        Attach the link to the pipeline.
+        :param pipeline: the pipeline
+        """
+        assert not self.pipeline, "Link already attached to a pipeline"
+
+        self.pipeline = pipeline
```
thiago-aixplain, line 136: Why do we need to keep an instance of pipeline inside every node and link?

aixplain/api/pipeline.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def add_link(self, link: Link) -> Link:
+        """
+        Add a link to the current pipeline.
+        :param link: the link
+        :return: the link
+        """
+        return link.attach(self)
+
+    def input(self, data: str = None, *args, **kwargs) -> Node:
```
thiago-aixplain, line 62: Can we add these methods which create a node into a method called `PipelineNodeFactory`?

aixplain/api/pipeline.py
```diff
@@ -0,0 +1,354 @@
...
+
+        :param data: the data to be uploaded
+        :param args: positional arguments
+        :param kwargs: keyword arguments
+        :return: the node
+        """
+        kwargs["data"] = data
+        return Input(self, *args, **kwargs)
+
+    def asset(self, assetId: str, *args, **kwargs) -> Node:
```
thiago-aixplain, line 79: Enable the user to also give the instance of a model:

```
assetId: Union[Model, Text]
```

aixplain/factories/script_factory.py
```diff
@@ -0,0 +1,31 @@
...
+
+import requests
+
+from aixplain.utils import config
+
+
+class ScriptFactory:
+
+    @classmethod
+    def upload_script(cls, script_path: str) -> Tuple[str, str]:
```
thiago-aixplain, line 13: For every factory, please create a corresponding `Asset` into aixplain.modules

thiago-aixplain (changes requested):

aixplain/api/base.py
kadirpekel: Because any node can be instantiated with a pipeline instance so that it can add itself to pipeline nodes:

All below the methods are valid for relating any node with a pipeline.
``` python
pipeline = Pipeline()
script = Script(script_path='myscript.py')
script.attach(pipeline)
```
or 
``` python
pipeline = Pipeline()
script = Script(script_path='myscript.py', pipeline=pipeline)
```
or
``` python
pipeline = Pipeline()
script = Script(script_path='myscript.py')
pipeline.add_node(script)
```

aixplain/api/pipeline.py
kadirpekel: I wouldn't do this as per our previous discussion on slack.

aixplain/factories/script_factory.py
kadirpekel: I only followed the current convention to upload a script using backend. What kind of asset we need here or i did not understand.  I referred to `FileFactory` while putting this and that factory does not use any kind of Asset. Do you mean sth else?

aixplain/api/pipeline.py
kadirpekel: This is sth i already planned (was certainly there in the first drafts) but missed to put it back. The same is also valid for Input node for the use of data assets. Let me check this by today. Will update accordingly

kadirpekel: Actually, indeed it was already there under `instance` param but still needs to be more convenient through. Will work on it.

kadirpekel: Ok this is done for assets and also Input while handling data from file. But we need to implement data asset, corpus or csv type of data usage which i need to understand how we could achieve that. But as a matter of having the first version we can proceed with as is and leave it as a next improvement.

thiago-aixplain (approved):

PR #216, opened 2024-08-05
Set model ID as a parameter for ModelTools

thiago-aixplain:


tests/functional/agent/data/agent_test_end2end.json
ikxplain, line 1: @thiago-aixplain there we are using a fixed model id, is this model public and shared through all envs?

ikxplain (approved):

PR #225, opened 2024-08-15
ENG-377 Added Validation Methods to ModelTool and PipeLine too

xainaz:
Added Validation methods to pipeline and model tools and called them in the create method.


aixplain/modules/agent/tool/model_tool.py
```diff
@@ -76,4 +76,11 @@ def __init__(
             model = model.id
         self.supplier = supplier
         self.model = model
-        self.function = function
\ No newline at end of file
+        self.function = function
+
+    def validate_model_tool():
```
thiago-aixplain, line 81: Since this method is inside a class, you need the `self` attribute. Moreover, since it is in the ModelTool class, I would simplify the name for: `def validate(self)`

aixplain/modules/agent/tool/model_tool.py
```diff
@@ -76,4 +76,11 @@ def __init__(
...
+        from aixplain.factories.model_factory import ModelFactory
+        try:
+            model = ModelFactory.get(model)
+        except:
+            raise Exception("Model Tool Unavailable.")
```
thiago-aixplain, line 86: Write a more detailed message. Something like:

f"Model Tool Unavailable. Make sure model '{model}' exists or you have access to it."

aixplain/modules/agent/tool/pipeline_tool.py
```diff
@@ -50,3 +50,11 @@ def __init__(
         if isinstance(pipeline, Pipeline):
             pipeline = pipeline.id
         self.pipeline = pipeline
+    
+
+    def validate_pipeline_tool():
```
thiago-aixplain, line 55: Same as model comments.

PR #230, opened 2024-08-16
Bug 73 sdk treat http errors

xainaz:


aixplain/modules/pipeline/asset.py
```diff
@@ -359,21 +319,31 @@ def run_async(
...
-            logging.info(
-                f"Result of request for {name}  - {r.status_code} - {resp}"
-            )
+            logging.info(f"Result of request for {name}  - {r.status_code} - {resp}")
 
             poll_url = resp["url"]
             response = {"status": "IN_PROGRESS", "url": poll_url}
         except Exception:
             response = {"status": "FAILED"}
+            if r.status_code == 401:
```
thiago-aixplain, line 333: Such conditions may be added at the begging of the `try` (after line 325). In this case, the method may return the following `response`:

```python
return {"status": "FAILED", "error": error_msg}
```

This comment is valid for all the modified running methods.

thiago-aixplain (changes requested):

PR #233, opened 2024-08-19
Adding supervisor/planning options into SDK

thiago-aixplain:


aixplain/factories/agent_factory/__init__.py
```diff
@@ -50,8 +50,44 @@ def create(
...
+        Args:
+            name (Text): name of the agent
+            llm_id (Text): aiXplain ID of the large language model to be used as agent.
+            tools (List[Tool], optional): list of tool for the agent. Defaults to [].
+            description (Text, optional): description of the agent role. Defaults to "".
+            api_key (Text, optional): team/user API key. Defaults to config.TEAM_API_KEY.
+            supplier (Union[Dict, Text, Supplier, int], optional): owner of the agent. Defaults to "aiXplain".
+            version (Optional[Text], optional): version of the agent. Defaults to None.
+            use_supervisor (bool, optional): flag to enable a multi-agent supervisor. Defaults to True.
+            supervisor_llm_id (Optional[Text], optional): aiXplain ID of the large language model to be used as the supervisor. Defaults to None.
```
lucas-aixplain, line 69: From what we discussed, I understood that we would not give the user the option to choose supervisor or planner llm id, we would just have this option internally in the communication between SDK and backend.

aixplain/factories/agent_factory/__init__.py
```diff
@@ -50,8 +50,35 @@ def create(
         api_key: Text = config.TEAM_API_KEY,
         supplier: Union[Dict, Text, Supplier, int] = "aiXplain",
         version: Optional[Text] = None,
+        use_supervisor: bool = True,
```
lucas-aixplain, line 53: Should we rename here to `use_orchestrator`? Since the community will also have the orchestrator in its name.

aixplain/factories/agent_factory/__init__.py
thiago-aixplain: Removed

aixplain/factories/agent_factory/__init__.py
thiago-aixplain: I removed this flag for now, since this option will always be in the architecture for now.

lucas-aixplain (approved):

PR #242, opened 2024-08-30
Eng 467 ai xplain sdk update finetune functional tests to cover all new finetunable models

xainaz:
Added createdAt attribute to model class, and implemented necessary changes in Model Factory. 
As for the finetune functional test, I replaced the changes with the original code that ran the test on the 3 models in data, since there were no new models in the past 3 weeks. 
2 models are passing ( llama2 7b, vectara) and 1 is failing : air

### DIFF
--- a/aixplain/modules/model/__init__.py
+++ b/aixplain/modules/model/__init__.py
@@ -239,11 +239,27 @@ class Model(Asset):
 
         resp = None
         try:
-            resp = r.json()
-            logging.info(f"Result of request for {name} - {r.status_code} - {resp}")
-
-            poll_url = resp["data"]
-            response = {"status": "IN_PROGRESS", "url": poll_url}
+            if 200 <= r.status_code < 300:
+                resp = r.json()
+                logging.info(f"Result of request for {name} - {r.status_code} - {resp}")
+                poll_url = resp["data"]
+                response = {"status": "IN_PROGRESS", "url": poll_url}
+            else:
+                if r.status_code == 401:
+                    error = "Unauthorized API key: Please verify the spelling of the API key and its current validity."
+                elif 460 <= r.status_code < 470:
+                    error = "Subscription-related error: Please ensure that your subscription is active and has not expired."
+                elif 470 <= r.status_code < 480:
+                    error = "Billing-related error: Please ensure you have enough credits to run this model. "
+                elif 480 <= r.status_code < 490:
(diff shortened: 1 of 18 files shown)


thiago-aixplain:
Where are the functional tests?

aixplain/modules/model/__init__.py
```diff
@@ -84,6 +85,7 @@ def __init__(
         self.backend_url = config.BACKEND_URL
         self.function = function
         self.is_subscribed = is_subscribed
+        self.createdAt = createdAt
 
     def to_dict(self) -> Dict:
         """Get the model info as a Dictionary
```
thiago-aixplain, line 91: Please check whether this `to_dict` method works with a datetime attribute.

tests/functional/finetune/finetune_functional_test.py
```diff
@@ -95,6 +109,13 @@ def test_end2end(run_input_map):
         assert len(result["details"][0]["metadata"]) > 0
     finetune_model.delete()
 
+def test_end2end(run_input_map, finetunable_llms_from_last_4_weeks):
+    if finetunable_llms_from_last_4_weeks:
+        for model in finetunable_llms_from_last_4_weeks:
+            run_finetune_test(model, run_input_map)
```
thiago-aixplain, line 115: You should not test all the models within the same method. You should call `test_end2end` for each model in `finetunable_llms_from_last_4_weeks`. Check how to use a fixture together with the python command `yield`. There is an example here: https://github.com/aixplain/aiXplain/blob/main/tests/functional/pipelines/designer_test.py#L17

thiago-aixplain (changes requested):

aixplain/modules/model/__init__.py
xainaz: It seems to work, this is what it returns:
{'id': '64e615671567f848804985e1', 'name': 'GPT2', 'supplier': <Supplier.OPENAI: {'id': 1777, 'name': 'OpenAI', 'code': 'openai'}>, 'additional_info': {'parameters': {'language': ['en']}}, 'createdAt': datetime.datetime(2023, 8, 30, 10, 53, 10, 827000, tzinfo=datetime.timezone.utc)}
I added createdAt to to_dict

thiago-aixplain (approved):

PR #243, opened 2024-08-30
 ENG-504: Make the agent architecture configurable

thiago-aixplain:


aixplain/factories/agent_factory/__init__.py
```diff
@@ -44,7 +44,7 @@ class AgentFactory:
     def create(
         cls,
         name: Text,
-        llm_id: Text,
+        llm_id: Text = "669a63646eb56306647e1091",
```
lucas-aixplain, line 47: Please add `Optional` in the method definition for all optional parameters e.g. `Optional[llm_id]`.

aixplain/factories/agent_factory/__init__.py
```diff
@@ -67,6 +67,8 @@ def create(
         Returns:
             Agent: created Agent
         """
+        # assert that at least one tool is provided
+        assert len(tools) > 0, "At least one tool must be provided."
```
lucas-aixplain, line 71: Do we need to force user to have at least one `ModelTool `or `PipelineTool`? I think if the user wants to use only serp api search we should let it do that.

aixplain/factories/agent_factory/__init__.py
thiago-aixplain: I think the use of `Optional` in type annotations in Python is intended to indicate that a value can either be of the specified type or `None`, no?. Since it has a not null value there is not need to use Optional. Please confirm.

lucas-aixplain: yes, you're correct. Didn't know that actually: https://docs.python.org/3/library/typing.html#typing.Optional

lucas-aixplain (approved):

PR #247, opened 2024-09-05
Eng 399 - Introducing Metric Nodes in Designer

kadirpekel:


aixplain/modules/pipeline/designer/nodes.py
```diff
@@ -129,6 +129,18 @@ def serialize(self) -> dict:
...
 
+class BareAssetInputs(Inputs):
+    pass
+
+
+class BareAssetOutputs(Outputs):
+    pass
+
+
+class BareAsset(AssetNode[BareAssetInputs, BareAssetOutputs]):
```
thiago-aixplain, line 140: What are these classes for?

thiago-aixplain (changes requested):
We may add the metric initializer here: https://github.com/aixplain/aiXplain/blob/main/aixplain/factories/pipeline_factory/utils.py#L57

aixplain/modules/pipeline/designer/nodes.py
kadirpekel: These classes are for concrete(bare) implementation of abstract AssetNode class. AssetNode is an abstract class in mind which expects typed input and output classes.

thiago-aixplain (approved):

PR #262, opened 2024-10-07
BUG-177: Fixed pipeline validation

kadirpekel:


aixplain/modules/pipeline/designer/pipeline.py
```diff
@@ -108,7 +109,7 @@ def validate_nodes(self):
 
         if not contains_input or not contains_output or not contains_asset:
             raise ValueError(
-                "Pipeline must contain at least one input, output and asset node"  # noqa
+                "Pipeline must contain at least one input, output and 'Outputable' node"  # noqa
```
thiago-aixplain, line 112: Explain what Outputable is. The user may not know.

thiago-aixplain (changes requested):

aixplain/modules/pipeline/designer/pipeline.py
kadirpekel: resolved, thanks.

thiago-aixplain (approved):

PR #275, opened 2024-10-11
Update model running endpoints from v1 to v2

thiago-aixplain:


aixplain/modules/model/__init__.py
```diff
@@ -204,21 +204,21 @@ def run(
...
         """
         start = time.time()
-        try:
-            response = self.run_async(data, name=name, parameters=parameters)
-            if response["status"] == "FAILED":
+        payload = build_payload(data=data, parameters=parameters)
+        url = f"{self.url}/api/v2/execute/{self.id}"
+        logging.debug(f"Model Run Sync: Start service for {name} - {url}")
+        response = call_run_endpoint(payload=payload, url=url, api_key=self.api_key)
+        if response["status"] == "IN_PROGRESS":
```
hadi-aix, line 211: Never assume that status will always be there. In case something horribly goes wrong, it wont. Better have the sdk handle it gracefully than crashing, no?

aixplain/modules/model/utils.py
```diff
@@ -0,0 +1,70 @@
...
+        logging.error(f"Error in request: {e}")
+        response = {
+            "status": "FAILED",
+            "completed": True,
+            "error_message": "Model Run: An error occurred while processing your request.",
+        }
+
+    if 200 <= r.status_code < 300:
+        logging.info(f"Result of request: {r.status_code} - {resp}")
+        status = resp.get("status", "IN_PROGRESS")
```
hadi-aix, line 46: same here, what is status is not there? It should be but better be safe and handle the case it is not and fail gracefully

aixplain/modules/model/utils.py
```diff
@@ -0,0 +1,70 @@
...
+            error = "Unauthorized API key: Please verify the spelling of the API key and its current validity."
+        elif 460 <= r.status_code < 470:
+            error = "Subscription-related error: Please ensure that your subscription is active and has not expired."
+        elif 470 <= r.status_code < 480:
+            error = "Billing-related error: Please ensure you have enough credits to run this model. "
+        elif 480 <= r.status_code < 490:
+            error = (
+                "Supplier-related error: Please ensure that the selected supplier provides the model you are trying to access."
+            )
+        elif 490 <= r.status_code < 500:
```
hadi-aix, line 63: Like we discussed, please return the exact error modelserving is returning unless you are doing this in another ticket

aixplain/modules/model/__init__.py
thiago-aixplain: In `call_run_endpoint` I make sure a status is always returned.

aixplain/modules/model/utils.py
thiago-aixplain: If the status is not there, it assigns `IN_PROGRESS`. In any case, I did a modification to return a response with FAILED status when no `data` is returned.

aixplain/modules/model/utils.py
thiago-aixplain: Done

hadi-aix (approved):

PR #279, opened 2024-10-18
Model Response Class

xainaz:


aixplain/modules/model/__init__.py
```diff
@@ -208,17 +209,28 @@ def run(
             response = self.run_async(data, name=name, parameters=parameters)
             if response["status"] == "FAILED":
                 end = time.time()
-                response["elapsed_time"] = end - start
-                return response
+                return ModelResponse(
+                    status="FAILED",
```
thiago-aixplain, line 213: Create an enumerator for the status with the values `FAILED`, `IN_PROGRESS` and `SUCCESS`. As a reference, follow this example: https://github.com/aixplain/aiXplain/blob/main/aixplain/enums/asset_status.py

aixplain/modules/model_response.py
```diff
@@ -0,0 +1,25 @@
...
+
+@dataclass
+class ModelResponse:
+    """Class for keeping track of an item in inventory."""
+
+    status: str
+    data: str
+    completed: bool
+    error_message: str
+
```
thiago-aixplain, line 12: Add used_credits and run_time as optional variables as well.

aixplain/modules/model_response.py
```diff
@@ -0,0 +1,25 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class ModelResponse:
+    """Class for keeping track of an item in inventory."""
+
+    status: str
```
thiago-aixplain, line 8: Use typing to set parameter type.

References:
- https://docs.python.org/3/library/typing.html
- https://github.com/aixplain/agentification/blob/development/agentification/utilities/models.py#L6

thiago-aixplain (changes requested):
Create the `model_response.py` script as `run_response.py` inside `aixplain/modules/model_response.py`

aixplain/modules/model/llm_model.py
```diff
@@ -142,13 +144,22 @@ def run(
             try:
                 poll_url = response["url"]
                 end = time.time()
-                response = self.sync_poll(poll_url, name=name, timeout=timeout, wait_time=wait_time)
+                return(self.sync_poll(poll_url, name=name, timeout=timeout, wait_time=wait_time))
```
thiago-aixplain, line 147: Fix the syntax here

thiago-aixplain (changes requested):

thiago-aixplain (approved):

PR #282, opened 2024-10-25
ENG-791: Pipeline Response Changes

xainaz:


aixplain/modules/pipeline/asset.py
```diff
@@ -140,10 +140,24 @@ def poll(self, poll_url: Text, name: Text = "pipeline_process") -> Dict:
...
             logging.info(f"Single Poll for Pipeline: Status of polling for {name} : {resp}")
-        except Exception:
-            resp = {"status": "FAILED"}
-        return resp
-
+            status = Status(resp.pop("status", "failed").lower())
+            print("resp in poll", resp)
+            response = PipelineResponse(
+                status=status,
+                error=resp.pop("error", ""),
```
kadirpekel, line 147: `None` as a default value for `error` would be much better.

aixplain/modules/pipeline_response.py
```diff
@@ -0,0 +1,54 @@
...
+        url: Optional[Text]="",
+        **kwargs,  
+    ):
+        self.status = status
+        self.error = error
+        self.elapsed_time = elapsed_time
+        self.data = data 
+        self.additional_fields = kwargs
+        self.url=url
+    def __getitem__(self, key: str) -> Any:
```
kadirpekel, line 23: using a auto formatter (eg Black) would help to avoid for such cases. For example a spacing is needed here to make the code more readable.

aixplain/modules/pipeline_response.py
```diff
@@ -0,0 +1,54 @@
+from dataclasses import dataclass
+from typing import Any, Optional, Dict, List, Union, Text
+from aixplain.enums import Status 
+
+@dataclass
+class PipelineResponse:
+    
+    def __init__(
```
kadirpekel, line 8: The use of dataclass here is not so correct. Dataclasses mostly designed to auto populate its attributes from the data fields defined. And also some other meta information also could be attached by mixing the use of both `InitVar` and the `__post_init__` function.

aixplain/modules/pipeline_response.py
```diff
@@ -0,0 +1,54 @@
...
+            fields.append(f"error={self.error}")
+        if self.elapsed_time is not None:
+            fields.append(f"elapsed_time={self.elapsed_time}")
+        if self.data:
+            fields.append(f"data={self.data}")
+        if self.additional_fields:
+            fields.extend([f"{k}={repr(v)}" for k, v in self.additional_fields.items()])
+        return f"PipelineResponse({', '.join(fields)})"
+
+    def __contains__(self, key: str) -> bool:
```
kadirpekel, line 49: above my comment regarding making this dict like object also applies here.

pipeline_test.py
```diff
@@ -0,0 +1,13 @@
...
+os.environ["MODELS_RUN_URL"] = "https://dev-models.aixplain.com/api/v1/execute"
+os.environ["TEAM_API_KEY"] = "65a701f4ddcc6be76a68996bd29255b85d1e335faa8eb0cdd5ebf7d857214e90"
+
+
+
+from aixplain.factories import PipelineFactory
+pipeline = PipelineFactory.get("66e40c5d8ee4eeb80a3aa819")
+result = pipeline.run("poem prompt")
+print("RESULT")
+print(result)
```
kadirpekel, line 13: We'd better remove redundant print statements, instead logging when necessary could be better approach

pipeline_test2.ipynb
kadirpekel, line 1: Is this file necessary to put under the repo? If yes, might there be better place to locate it?

tests/functional/pipelines/designer_test.py
```diff
@@ -117,13 +118,11 @@ def test_routing_pipeline(pipeline):
...
-
-    assert output["status"] == "SUCCESS"
-    assert output.get("data") is not None
-    assert len(output["data"]) > 0
-    assert output["data"][0].get("segments") is not None
-    assert len(output["data"][0]["segments"]) > 0
+    output = pipeline.run("This is a sample text!", **{"batchmode": False, "version": "3.0"})
+    print("output")
+    print(output)
+    print(output.status)
```
kadirpekel, line 124: many prints here as well, better to check the rest of the work in case any redundant print statements.

tests/functional/pipelines/run_test.py
```diff
@@ -20,156 +20,143 @@
...
-    assert response["status"] == "SUCCESS"
-    output = response["data"][0]
-    assert output["label"] == "Output 1"
-
-
-@pytest.mark.parametrize("version", ["2.0", "3.0"])
+from aixplain.enums.asset_status import AssetStatus
+
+
+# def test_list_pipelines():
```
kadirpekel, line 26: All these tests are disabled somehow. Is that by purpose?

aixplain/modules/pipeline_response.py
xainaz: Well we don't want to break any code that's already considering the response as a dictionary right? It's safer to keep both options I think.

tests/functional/pipelines/run_test.py
xainaz: The data is not being returned properly, so I commented some out to debug. Any idea why I could be getting success status, while no data is being returned?

pipeline_test2.ipynb
xainaz: Removed.

aixplain/modules/pipeline_response.py
```diff
@@ -0,0 +1,54 @@
...
+        **kwargs,  
+    ):
+        self.status = status
+        self.error = error
+        self.elapsed_time = elapsed_time
+        self.data = data 
+        self.additional_fields = kwargs
+        self.url=url
+    def __getitem__(self, key: str) -> Any:
+        if key in self.__dict__:
```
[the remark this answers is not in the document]
thiago-aixplain: The idea is to not break applications which use previous versions and assume a dictionary as response. Basically, legacy compatibility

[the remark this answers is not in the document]
kadirpekel: no i meant, instead of checking `__dict__` manually we can use sth like `getattr(self, key, None)`

aixplain/modules/pipeline_response.py
kadirpekel: Yes or no depends on your decision. In this version, you're almost using a regular class, no advantage in using dataclass here.

aixplain/modules/pipeline_response.py
kadirpekel: if it's the case, that's ok

tests/functional/pipelines/run_test.py
kadirpekel: I have really no idea on that.

aixplain/modules/pipeline/asset.py
```diff
@@ -152,41 +165,42 @@ def run(
...
-
-        Returns:
-            Dict: parsed output from pipeline
-        """
+    ) -> PipelineResponse:
         start = time.time()
         try:
             response = self.run_async(data, data_asset=data_asset, name=name, **kwargs)
-            if response["status"] == "FAILED":
+            if response["status"] == Status.FAILED:
```
kadirpekel, line 172: Here, are we sure the value of "status" in the matching enum type or a string only?

aixplain/modules/pipeline/asset.py
```diff
@@ -152,41 +165,42 @@ def run(
...
+                    status=Status.FAILED,
+                    error={"error": response.error, "status": "ERROR"},
+                    elapsed_time=end - start,
+                    **kwargs
+                )
             poll_url = response["url"]
+            polling_response = self.__polling(poll_url, name=name, timeout=timeout, wait_time=wait_time) 
             end = time.time()
-            response = self.__polling(poll_url, name=name, timeout=timeout, wait_time=wait_time)
+            status = Status(polling_response["status"])
```
kadirpekel, line 183: Regarding above my comment, for example here the value of "status" not seems to a be enum. Best to make sure of it.

aixplain/modules/pipeline/asset.py
xainaz: The enum makes sure that if it's passed in as string it turns it into enum object

aixplain/modules/pipeline/asset.py
xainaz: Okay I'll check again

aixplain/modules/pipeline/response.py
```diff
@@ -0,0 +1,55 @@
...
+        self.elapsed_time = elapsed_time
+        self.data = data
+        self.additional_fields = kwargs
+        self.url = url
+
+    def __getitem__(self, key: str) -> Any:
+        if key in self.__dict__:
+            return self.__dict__[key]
+
+        if self.additional_fields and key in self.additional_fields:
```
kadirpekel, line 28: related with above the comment, this could be moved under `getattr` overriding in conjuction with the use of `getattr` call in above if makes sense.

pipeline_test.py
kadirpekel: Above the code still exists. Still better to remove if not already.

tests/functional/pipelines/designer_test.py
kadirpekel: still better to remove those prints

tests/functional/pipelines/designer_test.py
```diff
@@ -152,14 +148,10 @@ def test_scripting_pipeline(pipeline):
...
-        version="2.0",
+        version="3.0",
     )
-
-    assert output["status"] == "SUCCESS"
-    assert output.get("data") is not None
-    assert len(output["data"]) > 0
-    assert output["data"][0].get("segments") is not None
-    assert len(output["data"][0]["segments"]) > 0
+    print(output.data)
```
kadirpekel, line 153: removing print statement in general is a good idea unless having a valid case

tests/functional/pipelines/designer_test.py
```diff
@@ -197,23 +189,22 @@ def test_decision_pipeline(pipeline):
...
 
-    output = pipeline.run("I feel so bad today!")
-
-    assert output["status"] == "SUCCESS"
+    output = pipeline.run(
+        "I feel so bad today!",
+        version="3.0",
+    )
+    print("output")
+    print(output)
```
kadirpekel, line 197: prints

aixplain/modules/pipeline/asset.py
```diff
@@ -157,38 +158,24 @@ def poll(self, poll_url: Text, name: Text = "pipeline_process") -> Dict:
...
-        if should_fallback:
-            logging.warning(
-                f"Pipeline Run Error: Failed to run pipeline {self.id} with version {version}. "
-                f"Trying with version {self.VERSION_2_0}."
+            if version == "v1":
+                return resp
+            status = ResponseStatus(resp.pop("status", "failed").lower())
+            response = PipelineResponse(
+                status=status,
+                error=resp.pop("error", "None"),
```
kadirpekel, line 166: don't know but should `error` be `"None"` as string when not available?

aixplain/modules/pipeline/asset.py
```diff
@@ -197,55 +184,51 @@ def run(
...
-                "status": "FAILED",
-                "error": error_message,
-                "elapsed_time": end - start,
-                "version": version,
-            }
+            if version == "v1":
+                return {"status": "failed", "error": error_message, "elapsed_time": end - start, **kwargs}
+            return PipelineResponse(
+                status=ResponseStatus.FAILED,
+                error={"error": error_message, "status": "ERROR"},
```
kadirpekel, line 228: here we see another type for `error` whereas it was string above. We should preserve the type consistency if we have no certain purpose.

kadirpekel (approved):

PR #284, opened 2024-10-28
Bug 206

kadirpekel:
`passthrough` parameters cause an automatic parameter population in the next node. This PR won't create a new one if there's already an existing one.


thiago-aixplain (approved):
Once you make sure the unit tests pass, please go ahead and merge it.

PR #294, opened 2024-10-30
Fix agent and team agent functional tests

lucas-aixplain:
Task: BUG-218

### DIFF
--- a/aixplain/modules/agent/__init__.py
+++ b/aixplain/modules/agent/__init__.py
@@ -249,12 +249,13 @@ class Agent(Model):
             headers = {"x-api-key": config.TEAM_API_KEY, "Content-Type": "application/json"}
             logging.debug(f"Start service for DELETE Agent  - {url} - {headers}")
             r = _request_with_retry("delete", url, headers=headers)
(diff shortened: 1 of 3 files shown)


aixplain/modules/agent/__init__.py
```diff
@@ -249,12 +249,13 @@ def delete(self) -> None:
...
             logging.debug(f"Start service for DELETE Agent  - {url} - {headers}")
             r = _request_with_retry("delete", url, headers=headers)
+            logging.debug(f"Result of request for DELETE Agent - {r.status_code}")
             if r.status_code != 200:
                 raise Exception()
         except Exception:
             try:
                 response_json = r.json()
-                message = f"Agent Deletion Error (HTTP {r.status_code}): {response_json.get('message')}."
+                message = f"Agent Deletion Error (HTTP {r.status_code}): {response_json.get('message').strip('{{}}')}."
```
thiago-aixplain, line 258: what about if there is not message? Wouldn't it be better something like:

```
response_json.get('message', "")
```

thiago-aixplain (approved):

PR #296, opened 2024-10-31
Eng 681

kadirpekel:


aixplain/factories/pipeline_factory/utils.py
```diff
@@ -82,6 +82,14 @@ def build_from_response(response: Dict, load_architecture: bool = False) -> Pipe
                         for output_param in node_json["outputValues"]
                         if output_param["code"] not in node.outputs
                     ]
+                if "customInputs" in node_json:
+                    for custom_input in node_json["customInputs"]:
+                        node.inputs.create_param(
+                            data_type=custom_input.get("dataType"),
+                            code=custom_input["code"],
+                            value=custom_input.get("value"),
```
thiago-aixplain, line 90: Is there always a value?

thiago-aixplain (approved):

aixplain/factories/pipeline_factory/utils.py
kadirpekel: No, `get` returns `None` if there's no value which is the case

PR #304, opened 2024-11-08
Bug 228 sdk always return model response when running model

xainaz:


aixplain/modules/model/response.py
```diff
@@ -9,7 +9,7 @@ class ModelResponse:
 
     def __init__(
         self,
-        status: ModelStatus,
+        status: AssetStatus,
```
lucas-aixplain, line 12: Here should be `ResponseStatus`, right?

tests/unit/llm_test.py
```diff
@@ -4,7 +4,7 @@
 
 load_dotenv()
 from aixplain.utils import config
-from aixplain.enums import ModelStatus
+from aixplain.enums.asset_status import AssetStatus
```
lucas-aixplain, line 7: Also, update the tests to check `ResponseStatus` instead of `AssetStatus`.

aixplain/modules/model/llm_model.py
```diff
@@ -29,7 +29,9 @@
 from aixplain.utils import config
 from typing import Union, Optional, List, Text, Dict
 from aixplain.modules.model.response import ModelResponse
-from aixplain.enums import ModelStatus
+from aixplain.enums.asset_status import AssetStatus
```
lucas-aixplain, line 32: @xainaz, do you have [pre-commit](https://pre-commit.com/) installed? Because since AssetStatus is not used here, our [flake8 hook](https://github.com/aixplain/aiXplain/blob/main/.pre-commit-config.yaml#L20-L25) should catch this and not let you commit without removing this line.

lucas-aixplain, line 32: @xainaz, do you have [pre-commit](https://pre-commit.com/) installed? Because since AssetStatus is not used here, our [flake8 hook](https://github.com/aixplain/aiXplain/blob/main/.pre-commit-config.yaml#L20-L25) should catch this and not let you commit without removing this line.

lucas-aixplain (changes requested):

lucas-aixplain (approved):

PR #316, opened 2024-11-19
Eng-1070: agent/team's output format

thiago-aixplain:


aixplain/modules/agent/response_format.py
```diff
@@ -0,0 +1,30 @@
...
+Date: February 21st 2024
+Description:
+    Asset Enum
+"""
+
+from enum import Enum
+from typing import Text
+
+
+class ResponseFormat(Text, Enum):
```
lucas-aixplain, line 28: In the agents repo, we're calling `OutputFormat`. Should we have the same name?

lucas-aixplain: Reminder to update from `ResponseFormat` to `OutputFormat`.

lucas-aixplain (approved):

PR #317, opened 2024-11-21
Validate prompt variables

thiago-aixplain:


aixplain/modules/agent/__init__.py
```diff
@@ -238,9 +238,24 @@ def run_async(
...
+            if isinstance(data, dict):
+                assert (
+                    variable in data or variable in parameters
+                ), f"Variable '{variable}' not found in data or parameters. This variable is required by the agent according to its description ('{self.description}')."
+                input_data[variable] = data.pop(variable) if variable in data else parameters.pop(variable)
+            else:
+                assert (
+                    variable in parameters
+                ), f"Variable '{variable}' not found in parameters. This variable is required by the agent according to its description ('{self.description}')."
+                input_data[variable] = parameters.pop(variable)
```
lucas-aixplain, line 254: We should add this piece of code to a method and call it here and in the team agent class. Also, we could test separately in the unit tests.

thiago-aixplain: done

lucas-aixplain (approved):

PR #318, opened 2024-11-22
Github Action Workflows

T-Alhussain:
Added unit and functional test workflows


.github/workflows/generate_pipeline.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install requests jinja2 pathlib aixplain
+
+      - name: Run generate.py to create pipeline.py
+        working-directory: aixplain/modules/pipeline
+        env:
+          BACKEND_URL: https://test-platform-api.aixplain.com
```
thiago-aixplain, line 30: These URLs may be updated. If it is too difficult to update with the one corresponding with the environtment, consider the production URL as the one to add here.

thiago-aixplain (approved):

PR #324, opened 2024-11-27
Eng 1129 ai xplain sdk caching functions

xainaz:


aixplain/enums/function.py
```diff
@@ -27,9 +27,53 @@
...
+def save_to_cache(functions, functions_input_output):
+    def convert_sets(obj):
+        if isinstance(obj, dict): return {k: convert_sets(v) for k, v in obj.items()}
+        elif isinstance(obj, list): return [convert_sets(i) for i in obj]
+        elif isinstance(obj, set): return list(obj)
+        else: return obj
+    cache_data = {"timestamp": time.time(), "functions": {"enum": {name: value for name, value in functions.__members__.items()}, "input_output": convert_sets(functions_input_output)}}
+    temp_file = CACHE_FILE + ".tmp"
+    try:
+        os.makedirs(os.path.dirname(CACHE_FILE), exist_ok=True); json_str = json.dumps(cache_data); logging.info(f"Serialized JSON size: {len(json_str)} bytes")
```
thiago-aixplain, line 48: Why everything in the same line?

aixplain/enums/function.py
```diff
@@ -27,9 +27,53 @@
...
+        with open(temp_file, "w") as f:
+            f.write(json_str); f.flush(); os.fsync(f.fileno())
+        os.replace(temp_file, CACHE_FILE); logging.info("Cache saved successfully.")
+    except Exception as e: logging.error(f"Failed to save cache: {e}")
 
+def convert_lists_to_sets(obj):
```
thiago-aixplain, line 54: Why do we need to do this conversion?

aixplain/enums/function.py
```diff
@@ -27,9 +27,53 @@
 from aixplain.utils.request_utils import _request_with_retry
 from enum import Enum
 from urllib.parse import urljoin
+import logging
+import time
+from datetime import datetime, timedelta
+import json
+import os
+
+CACHE_FILE = ".aixplain_cache/functions.json"
```
thiago-aixplain, line 36: add these constants into aixplain.utils.config

aixplain/enums/function.py
thiago-aixplain: why do we need to serialize the JSON in bytes? Couldn't you save it using:

```
import json
with open(CACHE, "w") as f:
   json.dump(dictionary, f, indent=4)
```

aixplain/enums/language.py
```diff
@@ -22,14 +22,59 @@
...
+    cache_data = {
+        "timestamp": time.time(),
+        "languages": languages,
+    }
+    temp_file = CACHE_FILE + ".tmp"
+    try:
+        os.makedirs(os.path.dirname(CACHE_FILE), exist_ok=True)
+        with open(temp_file, "w") as f:
+            json.dump(cache_data, f)
+            f.flush()
```
thiago-aixplain, line 48: why we need to flush?

aixplain/enums/language.py
```diff
@@ -22,14 +22,59 @@
...
 from enum import Enum
 from urllib.parse import urljoin
+from aixplain.utils import config
+from aixplain.utils.request_utils import _request_with_retry
 
+CACHE_FILE = ".aixplain_cache/languages.json"
+CACHE_DURATION = timedelta(hours=24)
+
+
+def save_to_cache(languages):
```
thiago-aixplain, line 38: Try to create a generic `save_to_cache` file that could be used across all files.

thiago-aixplain (changes requested):

aixplain/enums/function.py
xainaz: When I used dump it cut off abruptly for some reason, I changed to serialize and then right, that worked and the json files looked more readable too.

aixplain/enums/function.py
xainaz: I had a problem with parsing as I mentioned, I thought this might be related because JSON doesn’t support sets, so I changed the sets to lists when saving. I didn’t test it without this conversion after fixing the issue though. I will test removing the conversion, and if everything works fine without it, I’ll clean it up and remove this function.

aixplain/enums/language.py
xainaz: I added flush because the code was cutting off the JSON file while writing, and I thought it might help. flush makes sure everything in the program’s write buffer is sent to the file right away as far as I understand, so the file is fully written before anything else happens. Again this was part of my debugging process to ensure the data was saved correctly. I'll check out if it works properly without it, but out of curiosity isn't is usually recommended to have something like this when writing to external files regardless if the function is not running into any issues? Just to make sure that the sequence of actions runs smoothly

aixplain/enums/function.py
thiago-aixplain: Please investigate this deeply. The outcome is already a JSON as you can see [here](https://github.com/aixplain/aiXplain/blob/main/aixplain/enums/function.py#L44). My guess is that there is not need to change anything.

aixplain/enums/cache_utils.py
thiago-aixplain, line 1: Add this file in this folder: https://github.com/aixplain/aiXplain/tree/main/aixplain/utils

aixplain/enums/function.py
thiago-aixplain: Don't forget to add the CACHE_FILE into the right constant place.

thiago-aixplain (changes requested):

thiago-aixplain (approved):

PR #326, opened 2024-11-28
ENG-1153: Enabling custom python code tool

thiago-aixplain:
# Code Interpreter
```python
from aixplain.factories import AgentFactory
agent = AgentFactory.create(
    name="Python Developer",
    description="A Python developer agent. If you get an error from a tool, try to fix it.",
    tools=[AgentFactory.create_python_interpreter_tool()],
)
response = agent.run("Solve the equation $\\frac{v^2}{2} + 7v - 16 = 0$ to find the value of $v$.")
```

# Custom Code
```python
from aixplain.factories import AgentFactory

def main(a: int, b: int) -> int:
    """Add two numbers"""
    return a + b


agent = AgentFactory.create(
    name="Test Agent",
    tools=[
        AgentFactory.create_custom_python_code_tool(code=main),
    ],
    description="This is a test agent",
)
response = agent.run("How much is 2 + 2?")
```


tests/functional/agent/agent_functional_test.py
```diff
@@ -94,6 +94,30 @@ def test_end2end(run_input_map, delete_agents_and_team_agents):
     agent.delete()
 
 
+def test_python_interpreter_tool(delete_agents_and_team_agents):
```
lucas-aixplain, line 97: We should create a functional test for running with custom code as well, if it's not too costly.

lucas-aixplain (approved):

PR #349, opened 2024-12-23
ENG-1282: Set default pipeline engine version to 3.0 with fallback to…

kadirpekel:
… 2.0 in case of failure

### DIFF
--- a/aixplain/modules/pipeline/asset.py
+++ b/aixplain/modules/pipeline/asset.py
@@ -45,6 +45,9 @@ class Pipeline(Asset):
         **additional_info: Any additional Pipeline info to be saved
     """
 
+    VERSION_3_0 = "3.0"
+    VERSION_2_0 = "2.0"
+
     def __init__(
         self,
         id: Text,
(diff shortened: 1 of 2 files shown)


aixplain/modules/pipeline/asset.py
```diff
@@ -144,11 +153,44 @@ def poll(self, poll_url: Text, name: Text = "pipeline_process") -> Dict:
...
+        """
+        # If the version is not 3.0, no fallback is needed
+        if version != self.VERSION_3_0:
+            return False
+
+        should_fallback = False
+        if "status" not in response or response["status"] == "FAILED":
+            should_fallback = True
+        elif response["status"] == "SUCCESS" and (
+            "data" not in response or not response["data"]
```
thiago-aixplain, line 181: Should we also include when `len(response['data']) == 0`?

kadirpekel: `not response["data"]` already implies for empty lists also
```
if []:
    print("this wont happen")
```

thiago-aixplain (approved):

PR #355, opened 2025-01-03
ENG-1282 revert

thiago-aixplain:


aixplain/modules/pipeline/asset.py
```diff
@@ -215,15 +185,13 @@ def run(
         Returns:
             Dict: parsed output from pipeline
         """
-        version = version or self.VERSION_3_0
```
kadirpekel, line 218: There should be `VERSION_3_0` `VERSION_2_0` constants at somewhere if you also want to remove.

kadirpekel (approved):

PR #363, opened 2025-01-13
ENG-1393: index improvements

thiago-aixplain:


tests/functional/model/run_model_test.py
```diff
@@ -55,11 +55,11 @@ def test_run_async():
 
 
 def test_index_model():
-    from aixplain.modules.model.document_index import DocumentIndex
+    from aixplain.modules.model.record import Record
     from aixplain.factories import IndexFactory
 
     index_model = IndexFactory.create("test", "test")
-    index_model.add([DocumentIndex(value="Hello, world!", value_type="text", uri="", attributes={})])
+    index_model.upsert([Record(value="Hello, world!", value_type="text", uri="", attributes={})])
```
lucas-aixplain, line 62: We should also do a test where upsert method is used to update the index.

lucas-aixplain (approved):

PR #364, opened 2025-01-14
Prod-1534:  Utility Asset Improvements

ahmetgunduz:
- ENG-1442: Utility Asset Status added
- ENG-1443: Decorator based Utility Asset definition with any function name

### DIFF
--- a/aixplain/factories/model_factory/__init__.py
+++ b/aixplain/factories/model_factory/__init__.py
@@ -48,6 +48,7 @@ class ModelFactory:
         inputs: List[UtilityModelInput] = [],
         description: Optional[Text] = None,
         output_examples: Text = "",
+        api_key: Optional[Text] = None,
     ) -> UtilityModel:
         """Create a utility model
 
@@ -57,10 +58,11 @@ class ModelFactory:
             description (Text, optional): description of the model
             inputs (List[UtilityModelInput], optional): inputs of the model
             output_examples (Text, optional): output examples
-
+            api_key (Text, optional): Team API key. Defaults to None.
         Returns:
             UtilityModel: created utility model
         """
+        api_key = config.TEAM_API_KEY if api_key is None else api_key
         utility_model = UtilityModel(
             id="",
             name=name,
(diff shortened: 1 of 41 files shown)


aixplain/factories/model_factory/__init__.py
```diff
@@ -43,8 +43,8 @@ class ModelFactory:
     @classmethod
     def create_utility_model(
         cls,
-        name: Text,
-        code: Union[Text, Callable],
+        name: Optional[Text] = None,
+        code: Union[Text, Callable] = None,
```
thiago-aixplain, line 47: Can code be null here?

aixplain/modules/model/__init__.py
```diff
@@ -58,7 +58,7 @@ class Model(Asset):
     def __init__(
         self,
         id: Text,
-        name: Text,
+        name: Text = "",
```
thiago-aixplain, line 61: Name can not be an empty string. Are we doing some kind of validation to avoid this?

aixplain/modules/model/utility_model.py
```diff
@@ -113,16 +154,25 @@ def __init__(
...
+            try:
+                status = AssetStatus(status)
+            except Exception:
+                status = AssetStatus.DRAFT
+        self.status = status
 
     def validate(self):
         """Validate the Utility Model."""
         description = None
+        name = None
```
thiago-aixplain, line 167: Name can not be null, right?

aixplain/factories/model_factory/__init__.py
ahmetgunduz: you are right this is wrong. Fixed it

aixplain/modules/model/utility_model.py
ahmetgunduz: that is correct. I just initialize the name as null here but if it is set to empty string if not set properly

aixplain/factories/model_factory/__init__.py
ahmetgunduz: For the code it will check anyway when trying to initialize utility model.

PR #371, opened 2025-01-19
ENG-1440: refactoring factories to aiXplain SDK v2

thiago-aixplain:
Assets:
- api key
- corpus
- data
- dataset
- wallet

### DIFF
--- a/aixplain/v2/api_key.py
+++ b/aixplain/v2/api_key.py
@@ -1,0 +1,51 @@
+from .resource import (
+    BaseResource,
+    BareListParams,
+    BareGetParams,
+    GetResourceMixin,
+    ListResourceMixin,
+)
+from aixplain.factories import APIKeyFactory
+from aixplain.modules import APIKeyLimits, APIKeyUsageLimit
+from aixplain.utils import config
+from datetime import datetime
+from typing_extensions import Unpack
+from typing import Dict, List, Optional, Text, Union
+
+
+class APIKey(
+    BaseResource,
+    GetResourceMixin[BareGetParams, "APIKey"],
+    ListResourceMixin[BareListParams, "APIKey"],
+):
+    @classmethod
+    def get(cls, **kwargs: Unpack[BareGetParams]) -> "APIKey":
+        import aixplain.utils.config as config
+
+        api_key = kwargs.get("api_key", config.TEAM_API_KEY)
+        return APIKeyFactory.get(api_key=api_key)
+
+    @classmethod
+    def list(cls) -> Dict:
+        return APIKeyFactory.list()
+
+    @classmethod
+    def create(
+        cls,
+        name: Text,
+        budget: int,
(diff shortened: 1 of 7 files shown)


aixplain/v2/api_key.py
```diff
@@ -0,0 +1,51 @@
...
+
+        api_key = kwargs.get("api_key", config.TEAM_API_KEY)
+        return APIKeyFactory.get(api_key=api_key)
+
+    @classmethod
+    def list(cls) -> Dict:
+        return APIKeyFactory.list()
+
+    @classmethod
+    def create(
```
kadirpekel, line 33: We have `CreateResourceMixin` should be better applied and used here by also unpacking the related params.

aixplain/v2/corpus.py
```diff
@@ -0,0 +1,77 @@
...
+    @classmethod
+    def get(cls, **kwargs: Unpack[BareGetParams]) -> "Corpus":
+        return CorpusFactory.get(corpus_id=kwargs["id"])
+
+    @classmethod
+    def list(cls, **kwargs: Unpack[CorpusListParams]) -> List["Corpus"]:
+        return CorpusFactory.list(**kwargs)["results"]
+
+    @classmethod
+    def create(
```
kadirpekel, line 51: Again. We have `CreateResourceMixin` should be better applied.

aixplain/v2/dataset.py
```diff
@@ -0,0 +1,97 @@
...
+    data_type: Optional[DataType] = None
+    license: Optional[License] = None
+    page_number: int = 0
+    page_size: int = 20
+
+
+class Dataset(
+    BaseResource,
+    ListResourceMixin[DatasetListParams, "Dataset"],
+    GetResourceMixin[BareGetParams, "Dataset"],
```
kadirpekel, line 40: Again. We have `CreateResourceMixin` should be better applied.

aixplain/v2/finetune.py
```diff
@@ -7,10 +7,9 @@
     BareCreateParams,
 )
 
-if TYPE_CHECKING:
```
kadirpekel, line 10: Why did you modified here like so. Importing those modules directly will cause error because it will ask for environment variables to be set as in the legacy.

aixplain/v2/api_key.py
```diff
@@ -0,0 +1,51 @@
...
+    BaseResource,
+    GetResourceMixin[BareGetParams, "APIKey"],
+    ListResourceMixin[BareListParams, "APIKey"],
+):
+    @classmethod
+    def get(cls, **kwargs: Unpack[BareGetParams]) -> "APIKey":
+        import aixplain.utils.config as config
+
+        api_key = kwargs.get("api_key", config.TEAM_API_KEY)
+        return APIKeyFactory.get(api_key=api_key)
```
kadirpekel, line 26: We need to wrap the the returned objects with the Resource class itself where it needs across the PR.

PR #373, opened 2025-01-21
ENG-1291: Adding agents tasks

thiago-aixplain:


aixplain/modules/agent/agent_task.py
```diff
@@ -0,0 +1,23 @@
+from typing import List, Optional, Text
+
+
+class AgentTask:
+    def __init__(
+        self,
+        name: Text,
+        description: Text,
+        expected_output: Optional[Text] = None,
+        dependencies: Optional[List[Text]] = None,
```
ahmetgunduz, line 10: this should be also supporting AgentTask directly.

aixplain/modules/agent/agent_task.py
```diff
@@ -0,0 +1,23 @@
+from typing import List, Optional, Text
+
+
+class AgentTask:
```
ahmetgunduz, line 4: Why not we name it as Task, is there a reason for that?

ahmetgunduz (commented):
@Thiago left some minor comments

aixplain/modules/agent/agent_task.py
thiago-aixplain: Done

aixplain/modules/agent/agent_task.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(
+        self,
+        name: Text,
+        description: Text,
+        expected_output: Optional[Text] = None,
```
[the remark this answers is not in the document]
thiago-aixplain: I made this field required now.

aixplain/modules/agent/agent_task.py
lucas-aixplain: Are we using `AgentTask` or `Task`? In the PR it's still named `AgentTask`.

aixplain/modules/agent/__init__.py
```diff
@@ -73,6 +74,7 @@ def __init__(
         version: Optional[Text] = None,
         cost: Optional[Dict] = None,
         status: AssetStatus = AssetStatus.DRAFT,
+        tasks: Optional[List[AgentTask]] = None,
```
lucas-aixplain, line 77: Here we should initialize tasks the same way as in AgentFactory:
`tasks: List[AgentTask] = []`

tests/functional/team_agent/team_agent_functional_test.py
```diff
@@ -236,3 +236,35 @@ def test_add_remove_agents_from_team_agent(run_input_map, delete_agents_and_team
 
     team_agent.delete()
     new_agent.delete()
+
+
+def test_team_agent_tasks(delete_agents_and_team_agents):
```
lucas-aixplain, line 241: Can we have single agent with tasks and run it? If yes, we should create a functional test for that. If not, we should restrict that when creating the single agent.

aixplain/modules/agent/agent_task.py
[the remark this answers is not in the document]
ahmetgunduz: It still oprional here: https://github.com/aixplain/agentification/blob/42cdf1a6f81401594ce0517c38b2b16a91d09105/agentification/utilities/models.py#L143

aixplain/modules/agent/agent_task.py
thiago-aixplain: I decided to roll back to AgentTask. `Task` is too generic for our SDK which may have another Task concept in the future.

tests/functional/team_agent/team_agent_functional_test.py
thiago-aixplain: tasks in single agents are irrelevant. They are just used on teams.

aixplain/modules/agent/__init__.py
thiago-aixplain: Done

lucas-aixplain (approved):

PR #375, opened 2025-01-22
ENG-1484: Added an inspector Id to the team agent

OsujiCC:


yunsukim86 (changes requested):
Please add `use_mentalist` and `use_inspector` args in `create()`, and make sure `inspectorId` is set only when `use_inspector` is true.

Also raise an exception when `use_inspector` is true while `use_mentalist` is false.

Do not remove `use_mentalist_and_inspector` but keep it as legacy code. We can delete them later.

p.s. SDK calls a service to create an agent! Is our `development` branch of `agentification` running as a service now?

yunsukim86 (changes requested):
Please add `use_inspector` option and replace `mentalist_and_inspector_llm_id` with `inspector_llm_id`.

thiago-aixplain (approved):

PR #378, opened 2025-01-24
ENG-1494: Listing models by IDs

thiago-aixplain:


aixplain/factories/model_factory/utils.py
```diff
@@ -146,9 +147,49 @@ def get_assets_from_page(
...
         logging.error(error_message)
         raise Exception(error_message)
+
+
+def get_model_from_ids(model_ids: List[str], api_key: Optional[str] = None) -> List[Model]:
+    from aixplain.factories.model_factory.utils import create_model_from_response
+
+    resp = None
+    try:
+        url = urljoin(config.BACKEND_URL, f"sdk/models?ids={','.join(model_ids)}")
```
lucas-aixplain, line 168: Would an empty `model_ids` list work here? If not, we should handle this case.

lucas-aixplain (approved):

PR #384, opened 2025-02-03
ENG 1544: block mentalist without inspector for now and edit the test_model_repr function in unit test to make it pass.

OsujiCC:


aixplain/factories/team_agent_factory/__init__.py
```diff
@@ -99,8 +86,7 @@ def create(
             "description": description,
             "llmId": llm_id,
             "supervisorId": llm_id,
-            "plannerId": mentalist_llm_id,
-            "inspectorId": inspector_llm_id,
```
thiago-aixplain, line 103: Keep the inspector field

thiago-aixplain (changes requested):

thiago-aixplain (approved):

PR #387, opened 2025-02-04
ENG-1557: model params are now available for designer assets

kadirpekel:


aixplain/modules/pipeline/designer/nodes.py
```diff
@@ -108,18 +104,29 @@ def _auto_populate_params(self):
                 )
         else:
             for item in spec["params"]:
-                self.inputs.create_param(
+                if item["code"] not in self.inputs:
```
lucas-aixplain, line 107: So are we adding the function and model parameters here? Now to get the function parameters you can refer to FunctionParameters class [here](https://github.com/aixplain/aiXplain/blob/development/aixplain/enums/function.py#L70).

kadirpekel: Yes we're getting function parameters from spec obtained here
```
from aixplain.enums.function import FunctionInputOutput
spec = FunctionInputOutput[self.function]["spec"]
 ```
Does the location you advised makes difference?

lucas-aixplain: No, it's the same. So it's fine do that way.

lucas-aixplain (approved):

PR #389, opened 2025-02-05
ENG-1551 ai xplain sdk caching onboarded models pipelines and agents

xainaz:
Also added file lock, global expiry time in seconds is configurable as an environmental variable.


ahmetgunduz (changes requested):
@xainaz  Thanks for the PR! 

I have a main issue with the current iplementation. We have separate cache methods for models, pipelines and agents. This will make very hard for us to make any change in future. Having all into one class named AixplainCache object that has more modular desigh such that the utils methods and all repetitive methods for different asset types should be part of this class. 

Can you please redesign your codebase if this make sense to you. Otherwise we can discuss to find a way to make it more modular .

ahmetgunduz (changes requested):
@xainaz  Can you please add functional test for it? i approved the rest.

aixplain/enums/aixplain_cache.py
```diff
@@ -0,0 +1,117 @@
...
+    This class reduces code repetition and allows easier maintenance.
+    """
+
+    def __init__(self, asset_type: str, cache_filename: str):
+        """
+        Initialize the cache for a given asset type.
+
+        Args:
+            asset_type (str): Type of the asset (e.g., "models", "pipelines", "agents").
+            cache_filename (str): Filename for storing cached data.
```
thiago-aixplain, line 23: Make this field optional. If not provided, use the asset id as the filename

aixplain/enums/aixplain_cache.py
```diff
@@ -0,0 +1,117 @@
...
+                "created_at": asset.get("created_at", ""),
+                **asset,  # Include any extra fields
+            }
+            for asset in assets_data["items"]
+        }
+
+        return assets_enum, assets_details
+
+
+if __name__ == "__main__":
```
thiago-aixplain, line 105: Remove this main env

aixplain/modules/agent/__init__.py
```diff
@@ -80,6 +80,23 @@ def __init__(
         tasks: List[AgentTask] = [],
         **additional_info,
     ) -> None:
+        AgentCache = AixplainCache("agents", "agents")
```
thiago-aixplain, line 83: Add this code block after docstring

thiago-aixplain (changes requested):

aixplain/enums/aixplain_cache.py
thiago-aixplain: asset ID. One file for each asset in the proper folder.

aixplain/enums/aixplain_cache.py
```diff
@@ -0,0 +1,103 @@
...
+            cache_expiry (int, optional): Expiry time in seconds. Default is 24 hours.
+
+        Returns:
+            Tuple[Enum, Dict]: (Enum of asset IDs, Dictionary with asset details)
+        """
+        cached_data = load_from_cache(self.cache_file, self.lock_file)
+        if cached_data is not None:
+            return self.parse_assets(cached_data)
+
+        api_key = config.TEAM_API_KEY
```
kadirpekel, line 45: Better to move this fetching part to a dedicated function to make the structure more modular and testable.

aixplain/enums/aixplain_cache.py
```diff
@@ -0,0 +1,103 @@
...
+            logging.warning(f"No onboarded {self.asset_type} found.")
+            return Enum(self.asset_type.capitalize(), {}), {}
+
+        assets_enum = Enum(
+            self.asset_type.capitalize(),
+            {a["id"].upper().replace("-", "_"): a["id"] for a in assets_data["items"]},
+            type=str,
+        )
+
+        assets_details = {
```
kadirpekel, line 88: Better to use a dataclass, typed dict or pydantic model if present in the context instead of using arbitrary dictionaries for cache.

aixplain/enums/aixplain_cache.py
```diff
@@ -0,0 +1,103 @@
...
+        try:
+            response = _request_with_retry("get", url, headers=headers)
+            response.raise_for_status()
+            assets_data = response.json()
+        except Exception as e:
+            logging.error(f"Failed to fetch {self.asset_type} from API: {e}")
+            return Enum(self.asset_type.capitalize(), {}), {}
+
+        if "items" not in assets_data:
+            return Enum(self.asset_type.capitalize(), {}), {}
```
kadirpekel, line 59: Why are we using dynamic enums such like this? What is the benefit behind it?

aixplain/utils/cache_utils.py
```diff
@@ -2,26 +2,35 @@
...
-                return None
+def load_from_cache(cache_file, lock_file):
+    if os.path.exists(cache_file):
+        with FileLock(lock_file):
+            with open(cache_file, "r") as f:
+                cache_data = json.load(f)
+                if time.time() - cache_data["timestamp"] < int(get_cache_expiry()):
+                    return cache_data["data"]
+                else:
+                    return None
```
kadirpekel, line 35: If we know the cache is expired, shouldn't we cleanup the cache files and also lock files in case of need?

aixplain/enums/aixplain_cache.py
```diff
@@ -0,0 +1,103 @@
...
+from aixplain.utils.cache_utils import save_to_cache, load_from_cache, CACHE_FOLDER
+
+
+class AixplainCache:
+    """
+    A modular caching system to handle different asset types (Models, Pipelines, Agents).
+    This class reduces code repetition and allows easier maintenance.
+    """
+
+    def __init__(self, asset_type: str, cache_filename: Optional[str] = None):
```
kadirpekel, line 17: instead of expecting arbitrary strings here for `asset_type` we should either expect the real type of the asset conforming interface or at least a numeration.

aixplain/enums/aixplain_cache.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        Load assets from cache or fetch from backend if not cached.
+
+        Args:
+            cache_expiry (int, optional): Expiry time in seconds. Default is 24 hours.
+
+        Returns:
+            Tuple[Enum, Dict]: (Enum of asset IDs, Dictionary with asset details)
+        """
+        cached_data = load_from_cache(self.cache_file, self.lock_file)
```
kadirpekel, line 41: I believe that `load_from_cache` and `save_to_cache` function should be an integral part of this class if we're considering this class to be main caching logic for assets

aixplain/utils/cache_utils.py
```diff
@@ -2,26 +2,35 @@
...
+def save_to_cache(cache_file, data, lock_file):
     try:
         os.makedirs(os.path.dirname(cache_file), exist_ok=True)
-        with open(cache_file, "w") as f:
-            json.dump({"timestamp": time.time(), "data": data}, f)
+        with FileLock(lock_file):
+            with open(cache_file, "w") as f:
+                json.dump({"timestamp": time.time(), "data": data}, f)
     except Exception as e:
         logging.error(f"Failed to save cache to {cache_file}: {e}")
```
kadirpekel, line 24: why we're hiding the exception and log it? If we have no valid plan or purpose for preventing an exception to raise, we should let it raise to halt the program to proceed with an unwanted data flow.

aixplain/utils/cache_utils.py
kadirpekel, line 35: Where are we doing the house keeping? I mean if we here know that the cache expired, shouldn't we delete the waste?

aixplain/modules/model/__init__.py
```diff
@@ -91,9 +91,28 @@ def __init__(
             model_params (Dict, optional): parameters for the function.
             **additional_info: Any additional Model info to be saved
         """
+        ModelCache = AixplainCache("models", "models")
+        ModelEnum, ModelDetails = ModelCache.load_assets()
```
kadirpekel, line 95: Each time we're instantiating a model class (or any other asset using the same technique) will load the entire cache again and again which is quite inefficient . The entire cache should be instantiated once and loaded elsewhere independently from asset instantiation. Cache could be lazily populated at first demand or should be prepopulated when the library loaded.

aixplain/modules/pipeline/asset.py
```diff
@@ -74,6 +75,17 @@ def __init__(
             status (AssetStatus, optional): Pipeline status. Defaults to AssetStatus.DRAFT.
             **additional_info: Any additional Pipeline info to be saved
         """
+        PipelineCache = AixplainCache("pipelines", "pipelines")
```
kadirpekel, line 78: Having the same concerns above.

aixplain/modules/model/__init__.py
```diff
@@ -91,9 +91,28 @@ def __init__(
             model_params (Dict, optional): parameters for the function.
             **additional_info: Any additional Model info to be saved
         """
+        ModelCache = AixplainCache("models", "models")
```
kadirpekel, line 94: We should separate the caching logic and the core logic. Here he have the core logic of model asset and cache should be managed externally. In our case, as we have factory classes for each asset, those factory classes should be responsible to populate instances whether from cache for not.

aixplain/enums/aixplain_cache.py
```diff
@@ -0,0 +1,103 @@
...
+import logging
+from enum import Enum
+from urllib.parse import urljoin
+from typing import Dict, Optional, Tuple
+from aixplain.utils import config
+from aixplain.utils.request_utils import _request_with_retry
+from aixplain.utils.cache_utils import save_to_cache, load_from_cache, CACHE_FOLDER
+
+
+class AixplainCache:
```
kadirpekel, line 11: I won't rather such name containing the company name which is so generic. Such name like `AssetCache` would be more focused and pointing the purpose.

aixplain/enums/aixplain_cache.py
kadirpekel: Furthermore, instead of having individual `AssetCache` instances for each asset type, a more generic and centralized design could be considered if applicable.

aixplain/enums/aixplain_cache.py
xainaz: I initially used dynamic enums to generalize asset types, but switched to a static AssetType enum now — it's cleaner I think.

aixplain/enums/asset_cache.py
```diff
@@ -0,0 +1,174 @@
...
+    @classmethod
+    def from_dict(cls, data: Dict) -> "Model":
+        return cls(**data)
+
+@dataclass
+class Store(Generic[T]):
+    data: Dict[str, T]
+    expiry: int
+
+class AssetCache(Generic[T]):
```
kadirpekel, line 44: Why this modules stays in enums folder, what is logically related with enums?

aixplain/enums/asset_cache.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+T = TypeVar("T")
+
+# Constants
+CACHE_FOLDER = ".cache"
+DEFAULT_CACHE_EXPIRY = 86400
+
+@dataclass
+class Model:
```
kadirpekel, line 25: Are we using this `Model` class as it was sth part of the design document.

aixplain/enums/asset_cache.py
xainaz: We could move it to factories if that makes more sense?

aixplain/enums/asset_cache.py
xainaz: No no, I'll remove that, we'll import the Model module.

aixplain/enums/function.py
```diff
@@ -110,7 +135,11 @@ def __init__(self, input_params: Dict):
...
         for param_code, param_config in input_params.items():
-            self.parameters[param_code] = Parameter(name=param_code, required=param_config.get("required", False), value=None)
+            self.parameters[param_code] = Parameter(
+                name=param_code,
+                required=param_config.get("required", False),
+                value=None,
+            )
 
 
 Function, FunctionInputOutput = load_functions()
```
kadirpekel, line 145: Why we're caching functions as they're already inited and kept at module level where they're not already supposed initiated as per use.

aixplain/enums/function.py
```diff
@@ -34,23 +34,45 @@
 LOCK_FILE = f"{CACHE_FILE}.lock"
 
 
+class FunctionMetadata:
```
kadirpekel, line 37: What is the purpose of FunctionMetada as seeminly it has no value add rather than making a dictionary into plain class in untyped way. We'd better define the specs (fields) of FunctionMeta explicitly or otherwise we'd better not call this class as FunctionMeta because ise wildly generic. Maybe better call it a ObjectProxy class or similar.

aixplain/enums/function.py
kadirpekel: And also, apparently `load_functions` function does not return anything if i am not mistaken. how come we use it here?

aixplain/enums/language.py
```diff
@@ -25,39 +25,57 @@
...
 from aixplain.utils import config
 from aixplain.utils.request_utils import _request_with_retry
-from .asset_cache import AssetCache, CACHE_FOLDER
+from aixplain.utils.asset_cache import AssetCache
+import logging
 
-CACHE_FILE = f"{CACHE_FOLDER}/languages.json"
-LOCK_FILE = f"{CACHE_FILE}.lock"
+
+class LanguageMetadata:
```
kadirpekel, line 32: This also makes not sense as mentioned in FunctionMetadata. If we really need such a class, we'd better name it sth generic and use as general tool. But i would rather to make it explicit in term of spec by using a dataclass sort of thing.

aixplain/enums/language.py
```diff
@@ -25,39 +25,57 @@
...
+        for dialect in entry.dialects:
             dialect_label = "_".join(dialect["label"].split()).upper()
             dialect_value = dialect["value"]
+            languages[f"{label}_{dialect_label}"] = {
+                "language": language,
+                "dialect": dialect_value,
+            }
 
-            languages[language_label + "_" + dialect_label] = {"language": language, "dialect": dialect_value}
     return Enum("Language", languages, type=dict)
```
kadirpekel, line 81: Same consideration for languages. As they stay in module level during the lifetime of the app, why we would cache them? If the cache needed across different app instantiations, it's ok.

kadirpekel (approved):

thiago-aixplain (approved):

ahmetgunduz (approved):

PR #403, opened 2025-02-17
ENG-1616: Enable embedding models for air v2

thiago-aixplain:


aixplain/factories/index_factory.py
```diff
@@ -1,16 +1,18 @@
...
 
 
 class IndexFactory(ModelFactory):
     @classmethod
-    def create(cls, name: Text, description: Text) -> IndexModel:
+    def create(
+        cls, name: Text, description: Text, embedding_model: EmbeddingModel = EmbeddingModel.OPENAI_ADA002
+    ) -> IndexModel:
         """Create a new index collection"""
         model = cls.get("66eae6656eb56311f2595011")
```
lucas-aixplain, line 13: We should make this model id a constant variable as well.

aixplain/modules/model/index_model.py
```diff
@@ -50,14 +52,60 @@ def __init__(
         )
         self.url = config.MODELS_RUN_URL
         self.backend_url = config.BACKEND_URL
+        self.embedding_model = embedding_model
```
lucas-aixplain, line 55: Where is this used? We can add it to the dictionary representation of `IndexModel`.

results/inputs/67d9f3adc1699d26e2d62f4b-AI for Inventory Management Outreach Campaign.query.txt
lucas-aixplain, line 1: What is this file? If it's only a result file, it should be removed.

setup.cfg
```diff
@@ -1,3 +1,3 @@
 [metadata]
-description-file=README.md
+description_file=README.md
```
lucas-aixplain, line 2: This is fixed in another PR.

tests/functional/model/run_model_test.py
```diff
@@ -94,3 +107,58 @@ def test_llm_run_with_file():
...
+def test_index_model_with_image():
+    from aixplain.factories import IndexFactory
+    from aixplain.modules.model.record import Record
+    from uuid import uuid4
+
+    for index in IndexFactory.list()["results"]:
+        index.delete()
+
+    index_model = IndexFactory.create(
+        name=f"Image Index {uuid4()}", description="Index for images", embedding_model=EmbeddingModel.JINA_CLIP_V2_MULTIMODAL
```
lucas-aixplain, line 121: If this model works only for image input, should we check the type if this embedding was selected?

aixplain/factories/index_factory.py
thiago-aixplain: done

setup.cfg
thiago-aixplain: I had to do it since this branch is being used by a customer in a colab.

results/inputs/67d9f3adc1699d26e2d62f4b-AI for Inventory Management Outreach Campaign.query.txt
thiago-aixplain: Done

aixplain/modules/model/index_model.py
thiago-aixplain: This is used during indexing and is a required parameter to create an index. However, backend does not return it yet when getting the index model from the database. Backend team is working on it.

You mean added as a representation [here](https://github.com/aixplain/aiXplain/blob/main/aixplain/modules/model/__init__.py#L104)? I am not sure, since this is not an input parameter itself.

PR #416, opened 2025-02-26
'tatus ->status bug fixed '

ahmetgunduz:


tests/unit/agent/agent_test.py
```diff
@@ -110,7 +110,7 @@ def test_invalid_agent_name():
         AgentFactory.create(name="[Test]", description="", instructions="", tools=[], llm_id="6646261c6eb563165658bbb1")
     assert (
         str(exc_info.value)
-        == "Agent Creation Error: Agent name contains invalid characters. Only alphanumeric characters, spaces, hyphens, and brackets are allowed."
+        == "Agent Creation Error: Agent name contains invalid characters. Only alphanumeric characters, spaces, hyphens, and brackets are allowed."  # noqa: W503
```
thiago-aixplain, line 113: Remove the empty line at the end. No need for the qa comment

aixplain/modules/agent/agent_response.py
```diff
@@ -52,5 +52,5 @@ def to_dict(self) -> Dict[Text, Any]:
         return base_dict
 
     def __repr__(self) -> str:
-        fields = super().__repr__().strip("ModelResponse(").rstrip(")")
+        fields = super().__repr__()[len("ModelResponse(") : -1]  # noqa: E203
```
thiago-aixplain, line 55: Remove qa comment.

tests/unit/agent/agent_test.py
```diff
@@ -300,7 +300,7 @@ def test_update_success(mock_model_factory_get):
         # Capture warnings
         with pytest.warns(
             DeprecationWarning,
-            match="update\(\) is deprecated and will be removed in a future version. Please use save\(\) instead.",
+            match="update\(\) is deprecated and will be removed in a future version. Please use save\(\) instead.",  # noqa: W605
```
thiago-aixplain, line 303: Remove the qa comment.

Instead of match, is there a way to just check for a substring? Then we can remove the part that flake8 complains.

tests/unit/agent/agent_test.py
```diff
@@ -414,7 +414,7 @@ def test_run_variable_error():
         agent.run_async(data={"query": "Hello, how are you?"}, output_format=OutputFormat.MARKDOWN)
     assert (
         str(exc_info.value)
-        == "Variable 'target_language' not found in data or parameters. This variable is required by the agent according to its description ('Translate the input data into {target_language}')."
+        == "Variable 'target_language' not found in data or parameters. This variable is required by the agent according to its description ('Translate the input data into {target_language}')."  # noqa: W503
```
thiago-aixplain, line 417: Remove qa

aixplain/modules/agent/agent_response.py
ahmetgunduz: this one is problematic because the E203 error from flake8 indicates that there should not be any whitespace before the colon in a slice. However, black enforces a space around the colon in slices, which can lead to conflicts between the two tools.. We need to select one as reference. I propose to ignore E203 in flake.

ahmetgunduz: @thiago-aixplain thi

thiago-aixplain: to avoid this kind of problem, couldn't we change the solution to:

```
fields = super().__repr__().replace("ModelResponse(", "AgentResponse(")
```

thiago-aixplain (approved):

aixplain/modules/agent/agent_response.py
ahmetgunduz: let me check this

ahmetgunduz: Thanks thiago, this would be better I believe.

PR #421, opened 2025-03-06
Eng 1392 sqllite

ahmetgunduz:


tests/functional/agent/agent_functional_test.py
```diff
@@ -383,3 +382,27 @@ def test_sql_tool(delete_agents_and_team_agents, AgentFactory):
...
+    df.to_csv("test.csv", index=False)
+
+    tool = AgentFactory.create_sql_tool_from_csv(description="Test", csv_path="test.csv")
+    assert tool is not None
+    assert tool.description == "Test"
+    assert tool.database == "test.db"
+    assert tool.tables == ["test"]
+    assert tool.schema == 'CREATE TABLE test (\n                    "id" INTEGER, "name" TEXT, "value" REAL\n                )'
+    assert not tool.enable_commit  # must be False by default
+    assert tool.database == "test.db"
```
thiago-aixplain, line 403: Please run the agent with the SQL tool and check whether the response will be what is expected.

thiago-aixplain (changes requested):

tests/functional/agent/agent_functional_test.py
ahmetgunduz: Done

tests/functional/agent/agent_functional_test.py
```diff
@@ -347,14 +347,13 @@ def test_specific_model_parameters_e2e(tool_config):
     assert tool_used, "Tool was not used in execution"
 
 
-@pytest.mark.parametrize("AgentFactory", [AgentFactory, v2.Agent])
+@pytest.mark.parametrize("AgentFactory", [AgentFactory])
```
thiago-aixplain, line 350: put version 2 back.

aixplain/v2/agent.py
```diff
@@ -120,19 +118,3 @@ def create_custom_python_code_tool(cls, code: Union[str, Callable], description:
         from aixplain.factories import AgentFactory
 
         return AgentFactory.create_custom_python_code_tool(code=code, description=description)
-
```
thiago-aixplain, line 123: Sorry for the misunderstanding. Please add this piece of code back. This is actually a factory in version 2 of the SDK.

thiago-aixplain (changes requested):

aixplain/factories/agent_factory/__init__.py
```diff
@@ -193,7 +196,8 @@ def create_custom_python_code_tool(cls, code: Union[Text, Callable], description
     def create_sql_tool(
         cls,
         description: Text,
-        database: Text,
+        source: str,
+        source_type: str,
```
thiago-aixplain, line 200: Should we consider a default option?

Moreover can we do a Union on the typing hint with the Enumerators as well?

thiago-aixplain (approved):

aixplain/factories/agent_factory/__init__.py
ahmetgunduz: I do not think we need a default option here. Because there could be meny differnt ways to create sql tool

PR #442, opened 2025-03-19
ENG-1830: Setting default tool description

thiago-aixplain:
Model tool descriptions won't be sent as empty string anymore.

### DIFF
--- a/aixplain/factories/agent_factory/__init__.py
+++ b/aixplain/factories/agent_factory/__init__.py
@@ -136,7 +136,7 @@ class AgentFactory:
             raise Exception("Agent Onboarding Error: Please contact the administrators.")
 
         if 200 <= r.status_code < 300:
-            agent = build_agent(payload=response, api_key=api_key)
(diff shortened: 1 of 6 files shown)


aixplain/factories/agent_factory/__init__.py
```diff
@@ -136,7 +136,7 @@ def create(
             raise Exception("Agent Onboarding Error: Please contact the administrators.")
 
         if 200 <= r.status_code < 300:
-            agent = build_agent(payload=response, tools=tools, api_key=api_key)
+            agent = build_agent(payload=response, api_key=api_key)
```
lucas-aixplain, line 139: I don't think we should remove tools here because if we remove them, we'll load all the tools again, which will increase the agent creation time.

lucas-aixplain (changes requested):

aixplain/factories/agent_factory/__init__.py
thiago-aixplain: Changed according to suggestion

PR #443, opened 2025-03-20
ENG-1789: Add multiple index backbones support

basitanees:
Onboard Vectara, GraphRAG and ZeroEntropy

### DIFF
--- a/aixplain/enums/index_stores.py
+++ b/aixplain/enums/index_stores.py
@@ -5,6 +5,7 @@ class IndexStores(Enum):
     AIR = {"name": "air", "id": "66eae6656eb56311f2595011"}
     VECTARA = {"name": "vectara", "id": "655e20f46eb563062a1aa301"}
     GRAPHRAG = {"name": "graphrag", "id": "67dd6d487cbf0a57cf4b72f3"}
+    ZERO_ENTROPY = {"name": "zeroentropy", "id": "6807949168e47e7844c1f0c5"}
 
     def __str__(self):
         return self.value["name"]
--- a/aixplain/factories/index_factory/utils.py
+++ b/aixplain/factories/index_factory/utils.py
@@ -42,10 +42,11 @@ class VectaraParams(BaseIndexParams):
 
 
 class ZeroEntropyParams(BaseIndexParams):
-    _id: ClassVar[str] = ""
+    _id: ClassVar[str] = IndexStores.ZERO_ENTROPY.get_model_id()
 
-    def __init__(self, **kwargs):
-        raise ValueError("ZeroEntropy is not supported yet")
+    @property
+    def id(self) -> str:
+        return self._id
 
 
 class AirParams(BaseIndexParamsWithEmbeddingModel):
--- a/tests/functional/model/run_model_test.py
+++ b/tests/functional/model/run_model_test.py
@@ -8,7 +8,7 @@ from aixplain.factories import ModelFactory
 from aixplain.modules import LLM
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
-from aixplain.factories.index_factory.utils import AirParams, VectaraParams, GraphRAGParams
+from aixplain.factories.index_factory.utils import AirParams, VectaraParams, GraphRAGParams, ZeroEntropyParams
 
 
 def pytest_generate_tests(metafunc):
@@ -92,6 +92,7 @@ def run_index_model(index_model):
     "embedding_model,supplier_params",
     [
         pytest.param(None, VectaraParams, id="VECTARA"),
+        pytest.param(None, ZeroEntropyParams, id="ZERO_ENTROPY"),
         pytest.param(EmbeddingModel.OPENAI_ADA002, GraphRAGParams, id="GRAPHRAG"),
         pytest.param(EmbeddingModel.OPENAI_ADA002, AirParams, id="AIR - OpenAI Ada 002"),
         pytest.param(EmbeddingModel.SNOWFLAKE_ARCTIC_EMBED_M_LONG, AirParams, id="AIR - Snowflake Arctic Embed M Long"),


aixplain/enums/index_stores.py
```diff
@@ -0,0 +1,14 @@
+from enum import Enum
+
+
+class IndexStores(Enum):
+    AIR = {"name": "air", "id": "66eae6656eb56311f2595011"}
+    VECTARA = {"name": "vectara", "id": "655e20f46eb563062a1aa301"}
+    GRAPHRAG = {"name": "graphrag", "id": "67dd6d487cbf0a57cf4b72f3"}
+    # ZERO_ENTROPY = {"name": "zero_entropy", "id": ""}
```
thiago-aixplain, line 8: Remove comments

thiago-aixplain (changes requested):
Please write the unit tests [here](https://github.com/aixplain/aiXplain/blob/main/tests/unit/index_model_test.py) and the functional ones [here](https://github.com/aixplain/aiXplain/blob/main/tests/functional/model/run_model_test.py)

aixplain/factories/index_factory/utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+class ZeroEntropyParams(BaseIndexParams):
+    pass
+
+class AirParams(BaseIndexParams):
+    embedding_model: Optional[EmbeddingModel] = EmbeddingModel.OPENAI_ADA002 # should allow all embedding model ids as this is not very scalable
+
+class GraphRAGParams(BaseIndexParams):
+    embedding_model: Optional[EmbeddingModel] = EmbeddingModel.OPENAI_ADA002
+    llm_model: Optional[Text] = "669a63646eb56306647e1091" # Gpt-4o-mini
```
kadirpekel, line 42: Moving this id to a constant or to an existing enum would be better.

aixplain/factories/index_factory/utils.py
```diff
@@ -0,0 +1,39 @@
...
+    description: Optional[Text] = ""
+
+class VectaraParams(BaseIndexParams):
+    pass
+
+class ZeroEntropyParams(BaseIndexParams):
+    pass
+
+class AirParams(BaseIndexParams):
+    embedding_model: Optional[EmbeddingModel] = EmbeddingModel.OPENAI_ADA002 # should allow all embedding model ids as this is not very scalable
```
kadirpekel: I would rather a structure such as below, please consider it if applicable.
```
class BaseIndexParams(BaseModel):
    ...
    id: str
    model: Optional[EmbeddingModel] = None

class AirParams(BaseIndexParams):
    id: str = IndexStores.AIR.get_model_id()
    model: Optional[EmbeddingModel] = EmbeddingModel.OPENAI_ADA002 
    ...

class GraphRAGParams(BaseIndexParams):
    id: str = IndexStores.GRAPHRAG.get_model_id()
    model: Optional[EmbeddingModel] = EmbeddingModel.OPENAI_ADA002
    ...


class IndexFactory(ModelFactory):

    @classmethod
    def create(cls, params: BaseIndexParams) -> IndexModel:
        """Create a new index collection"""
        model = cls.get(params.model_id)
        payload = params.model_dump(exclude_none=True)

        response = model.run(data=payload)
        if response.status == ResponseStatus.SUCCESS:
            model_id = response.data
```

basitanees: Sure, let me try

aixplain/factories/index_factory/__init__.py
```diff
@@ -1,20 +1,42 @@
...
-from aixplain.enums import EmbeddingModel, Function, ResponseStatus, SortBy, SortOrder, OwnershipType, Supplier
-from typing import Optional, Text, Union, List, Tuple
-
-AIR_MODEL_ID = "66eae6656eb56311f2595011"
+from aixplain.enums import Function, ResponseStatus, SortBy, SortOrder, OwnershipType, Supplier
+from typing import Text, Union, List, Tuple, Optional
+from aixplain.factories.index_factory.utils import BaseIndexParams
 
 
 class IndexFactory(ModelFactory):
```
kadirpekel, line 31: You may consider using generics here

```
T = TypeVar("T", bound= BaseIndexParams)

class IndexFactory(ModelFactory, Generic[T]):
    def create(cls, params: T) -> IndexModel:
        ...

```

basitanees:
> Please write the unit tests [here](https://github.com/aixplain/aiXplain/blob/main/tests/unit/index_model_test.py) and the functional ones [here](https://github.com/aixplain/aiXplain/blob/main/tests/functional/model/run_model_test.py)

@thiago-aixplain Since, I only changed the IndexFactory create method, I don't think I need to write any unit tests using IndexModel.
Regarding the functional test, the current update will test it for aIR, but not for Vectara and GraphRAG. Do I need to write all tests for Vectara and GraphRAG before we can merge this, or can we do that at a later time as well? I am hoping to merge this soon so we could start using and testing with the SDK ourselves.

aixplain/enums/embedding_model.py
```diff
@@ -20,7 +20,7 @@
 from enum import Enum
 
 
-class EmbeddingModel(Enum):
+class EmbeddingModel(str, Enum):
```
thiago-aixplain, line 23: why do we need that?

tests/functional/model/run_model_test.py
```diff
@@ -106,24 +90,113 @@ def test_index_model(embedding_model):
...
     "embedding_model",
     [
+        pytest.param(EmbeddingModel.OPENAI_ADA002, id="OpenAI Ada 002"),
         pytest.param(EmbeddingModel.SNOWFLAKE_ARCTIC_EMBED_M_LONG, id="Snowflake Arctic Embed M Long"),
+        pytest.param(EmbeddingModel.SNOWFLAKE_ARCTIC_EMBED_L_V2_0, id="Snowflake Arctic Embed L v2.0"),
+        pytest.param(EmbeddingModel.MULTILINGUAL_E5_LARGE, id="Multilingual E5 Large"),
+        pytest.param(EmbeddingModel.BGE_M3, id="BGE M3"),
+    ],
+)
+def test_index_model_air(embedding_model):
```
thiago-aixplain, line 100: Can we parametrize the index store in a unique `text_index_model` function?

tests/functional/model/run_model_test.py
```diff
@@ -106,24 +90,113 @@ def test_index_model(embedding_model):
...
         pytest.param(EmbeddingModel.OPENAI_ADA002, id="OpenAI Ada 002"),
+        pytest.param(EmbeddingModel.SNOWFLAKE_ARCTIC_EMBED_M_LONG, id="Snowflake Arctic Embed M Long"),
         pytest.param(EmbeddingModel.SNOWFLAKE_ARCTIC_EMBED_L_V2_0, id="Snowflake Arctic Embed L v2.0"),
         pytest.param(EmbeddingModel.JINA_CLIP_V2_MULTIMODAL, id="Jina Clip v2 Multimodal"),
         pytest.param(EmbeddingModel.MULTILINGUAL_E5_LARGE, id="Multilingual E5 Large"),
         pytest.param(EmbeddingModel.BGE_M3, id="BGE M3"),
     ],
 )
-def test_index_model_with_filter(embedding_model):
+def test_index_model_air_with_filter(embedding_model):
```
thiago-aixplain, line 160: Can we parametrize index_stores on a `test_index_model_with_filter` function?

thiago-aixplain (changes requested):

aixplain/factories/index_factory/utils.py
thiago-aixplain: I suggest we do not this to do not break backward compatibility. We would have to change all documentation and colabs.

aixplain/enums/embedding_model.py
basitanees: If we dont do this, then we have to explicitly get the value using .value property (in case we pass an instance of EmbeddingModel), or using direct value if we pass the model id directly, like the following:
`
data.pop("embedding_model").value if isinstance(self.embedding_model, EmbeddingModel) else data.pop("embedding_model")
`

By inheriting from str, we could directly use 
`
data.pop("embedding_model")
`
for both cases.

aixplain/factories/index_factory/utils.py
basitanees: Could we keep the original function and create a new function with a different name for this then?
Or we could add a parameter called "index_type" and validate the params accordingly? (However, this does not seems to be as robust)

basitanees: Other option for backward compatibility would be to have kwargs as input, along with name, description and index type. However, that is not very intuitive as users will not know what args to input for which index type and we would still have to document it properly for each case.

basitanees: Pushed an update

basitanees:
Hi @Thiago This looks good.
For the tests, they will fail for Vectara and GraphRAG. For vectara, you cannot create a document with the same document id (unless you have deleted it first)
For GraphRAG, you cannot retrieve or delete a document.

thiago-aixplain (approved):

PR #449, opened 2025-03-24
ENG-1842: Utility asset status set to draft as default

ahmetgunduz:
- integrated with backend
- lifecycle of utilities are same as agents
- added test methods for status check and overwriting draft utility

### DIFF
--- a/aixplain/modules/model/utility_model.py
+++ b/aixplain/modules/model/utility_model.py
@@ -43,17 +43,20 @@ class UtilityModelInput:
     def to_dict(self):
         return {"name": self.name, "description": self.description, "type": self.type.value}
 
+
 # Tool decorator
-def utility_tool(name: Text, description: Text, inputs: List[UtilityModelInput] = None, output_examples: Text = "", status = AssetStatus.DRAFT):
+def utility_tool(
+    name: Text, description: Text, inputs: List[UtilityModelInput] = None, output_examples: Text = "", status=AssetStatus.DRAFT
+):
     """Decorator for utility tool functions
-    
+
     Args:
         name: Name of the utility tool
         description: Description of what the utility tool does
         inputs: List of input parameters, must be UtilityModelInput objects
         output_examples: Examples of expected outputs
         status: Asset status
-        
+
     Raises:
         ValueError: If name or description is empty
         TypeError: If inputs contains non-UtilityModelInput objects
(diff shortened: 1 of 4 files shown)


aixplain/modules/model/utility_model.py
```diff
@@ -116,7 +120,7 @@ def __init__(
         function: Optional[Function] = None,
         is_subscribed: bool = False,
         cost: Optional[Dict] = None,
-        status: AssetStatus = AssetStatus.ONBOARDED,# TODO: change to draft when we have the backend ready
+        status: AssetStatus = AssetStatus.DRAFT,
```
thiago-aixplain, line 123: Should we add a warning to inform the users that the utility model will expire after 24 hours if they do not deploy it?

tests/functional/model/run_utility_model_test.py
```diff
@@ -145,3 +150,61 @@ def test_string(input: str):
...
+        assert utility_model.inputs[1].type == DataType.TEXT
+
+        # try to reinitialize the model
+        utility_model = ModelFactory.create_utility_model(
+            name="Location Utility Test",
+            code=get_user_location,
+        )
+
+        # deploy the model
+        utility_model.deploy()
```
thiago-aixplain, line 193: assert the status before (DRAFT) and after (ONBOARDED) the deploy.

thiago-aixplain (changes requested):

aixplain/modules/model/utility_model.py
ahmetgunduz: Makes sense, do we have this in agents as well?

aixplain/modules/model/utility_model.py
```diff
@@ -161,8 +169,23 @@ def __init__(
...
+        if status == AssetStatus.DRAFT:
+            warnings.warn(
+                "WARNING: Non-deployed utility models (status=DRAFT) will expire after 24 hours after creation. "
+                "Use .deploy() method to make the model permanent.",
+                UserWarning,
+            )
+
     def validate(self):
         """Validate the Utility Model."""
+        if self.status == AssetStatus.DRAFT:
```
thiago-aixplain, line 181: Perhaps just leave on __init__?

thiago-aixplain (approved):

PR #455, opened 2025-03-25
ENG-1862:Added status to Tools and deployment check for Agent and TeamAgent

ahmetgunduz:


aixplain/modules/team_agent/__init__.py
```diff
@@ -406,17 +406,28 @@ def update(self) -> None:
             error_msg = f"Team Agent Update Error (HTTP {r.status_code}): {resp}"
             raise Exception(error_msg)
 
+    def _get_not_onboarded_agents(self, agents: List[Agent]) -> List[Agent]:
+        return [agent for agent in agents if agent.status != AssetStatus.ONBOARDED]
+
+    def _is_ready_to_deploy(self) -> bool:
```
kadirpekel, line 412: Name suggests returning a predicate but instead it raises exceptions which is also valid but with this naming it's misleading. Name could be sth like `validate_readiness`, `check_readiness` etc.

aixplain/modules/team_agent/__init__.py
```diff
@@ -406,17 +406,28 @@ def update(self) -> None:
             error_msg = f"Team Agent Update Error (HTTP {r.status_code}): {resp}"
             raise Exception(error_msg)
 
+    def _get_not_onboarded_agents(self, agents: List[Agent]) -> List[Agent]:
```
kadirpekel, line 409: Since a fetched list already been provided to this function, i would rather such a name `_filter_not_onboarded_agents`  or similar.

aixplain/modules/agent/__init__.py
```diff
@@ -202,6 +199,8 @@ def run(
             poll_url = response["url"]
             end = time.time()
             result = self.sync_poll(poll_url, name=name, timeout=timeout, wait_time=wait_time)
+            # if result.status == ResponseStatus.FAILED:
```
thiago-aixplain, line 202: Remove the comment

aixplain/modules/agent/__init__.py
```diff
@@ -409,11 +408,5 @@ def save(self) -> None:
         """Save the Agent."""
         self.update()
```
thiago-aixplain, line 410: Don't we have a deploy method anymore?

ahmetgunduz: It is now in mixins.py, and this class inherits from it

thiago-aixplain (approved):

PR #456, opened 2025-03-25
ENG-1832 add air 2 functionalities to the sdk

Muhammad-Elmallah:


thiago-aixplain (changes requested):
Please add unit tests [here](https://github.com/aixplain/aiXplain/blob/main/tests/unit/index_model_test.py) and functional test [here](https://github.com/aixplain/aiXplain/blob/main/tests/functional/model/run_model_test.py) for the new functionalities.

thiago-aixplain (approved):

PR #466, opened 2025-03-31
ENG-1899: Fixing test model ids test

thiago-aixplain:


tests/functional/general_assets/asset_functional_test.py
```diff
@@ -93,7 +93,7 @@ def test_model_supplier(ModelFactory):
 
 
 def test_model_ids():
-    model_ids = ["674728f51ed8e18fd8a1383f", "674728f51ed8e18fd8a1383c"]
+    model_ids = ["674728f51ed8e18fd8a1383f", "669a63646eb56306647e1091"]
```
lucas-aixplain, line 96: Could we have this list set as pytest params alongside their model names? That way, we'd know which model each ID refers to.

PR #476, opened 2025-04-02
ENG-1920: added sentry cred

xainaz:


thiago-aixplain (changes requested):
Add sentry dependency [here](https://github.com/aixplain/aiXplain/blob/main/pyproject.toml#L69)

thiago-aixplain (approved):

PR #485, opened 2025-04-10
ENG-1978: Adding instructions to teams

thiago-aixplain:


aixplain/factories/team_agent_factory/__init__.py
```diff
@@ -139,6 +141,7 @@ def create(
             "supplier": supplier,
             "version": version,
             "status": "draft",
+            "role": instructions,
```
yunsukim86, line 144: Why is it `role` instead of `instructions`?

yunsukim86 (commented):
What is the difference to `description`?

aixplain/factories/team_agent_factory/__init__.py
thiago-aixplain: Initially the field to set instructions to an agent/team was called `role`. Product asked to change to `instructions`. We changes only in the SDK though.

thiago-aixplain:
> What is the difference to `description`?

`description` is a summary of the functionality of agent/team does to depict in the marketplace. `instructions` is actually what is added in the LLM prompt.

yunsukim86 (approved):
lgtm

Please put the details of description/instructions in the docstring.

PR #495, opened 2025-04-17
ENG-2007: Fix agent and team agent parametrized functional test - Sync with dev

lucas-aixplain:


tests/functional/agent/agent_functional_test.py
```diff
@@ -327,6 +328,7 @@ def test_specific_model_parameters_e2e(tool_config):
         tools=[tool],
         llm_id="6626a3a8c8f1d089790cf5a2",  # Using LLM ID from test data
     )
+    agent.deploy()
```
thiago-aixplain, line 331: Why do we have to deploy?

lucas-aixplain: Without deploying, the second test is failing. The agent is not being overwritten, so it uses the search instead of the translation tool. Is that expected?

thiago-aixplain (approved):

PR #501, opened 2025-04-24
ENG-1836: Set name of tools on the SDK

thiago-aixplain:


aixplain/modules/agent/tool/model_tool.py
```diff
@@ -28,6 +28,30 @@
 from aixplain.modules.model import Model
 
 
+def set_tool_name(function: Function, supplier: Supplier = None, model: Model = None) -> Text:
```
ahmetgunduz, line 31: what will the name will be if it is an utility ?

aixplain/modules/agent/tool/custom_python_code_tool.py
```diff
@@ -21,17 +21,19 @@
...
 class CustomPythonCodeTool(Tool):
     """Custom Python Code Tool"""
 
-    def __init__(self, code: Union[Text, Callable], description: Text = "", **additional_info) -> None:
+    def __init__(
+        self, code: Union[Text, Callable], description: Text = "", name: Optional[Text] = None, **additional_info
+    ) -> None:
         """Custom Python Code Tool"""
-        super().__init__(name="Custom Python Code", description=description, **additional_info)
+        super().__init__(name=name or "", description=description, **additional_info)
```
ahmetgunduz, line 36: # default should be not empty string
name = name or "Custom Python Code Tool"
super().__init__(name=name, description=description, **additional_info)

aixplain/modules/agent/tool/pipeline_tool.py
```diff
@@ -65,7 +64,18 @@ def to_dict(self):
...
+            pipeline_obj = self.pipeline
+        else:
+            try:
+                pipeline_obj = PipelineFactory.get(self.pipeline, api_key=self.api_key)
+            except Exception:
+                raise Exception(
+                    f"Pipeline Tool Unavailable. Make sure Pipeline '{self.pipeline}' exists or you have access to it."
+                )
+
+        if self.name.strip() == "":
```
ahmetgunduz, line 77: if not self.name:
```suggestion
        if self.name.strip() == "":
```

aixplain/modules/agent/tool/python_interpreter_tool.py
ahmetgunduz, line 1: lets add a name param to thi tool as well to keep the same structure even though it is internal.

aixplain/modules/agent/__init__.py
```diff
@@ -127,11 +127,18 @@ def _validate(self) -> None:
...
+            tool_name = None
             if isinstance(tool, Tool):
-                tool.validate()
+                tool_name = tool.name
             elif isinstance(tool, Model):
                 assert not isinstance(tool, Agent), "Agent cannot contain another Agent."
+                tool_name = tool.name
+            assert (
+                tool_name not in tool_names
+            ), f"Agent Creation Error: Tool name '{tool_name}' is already used by another tool. Make sure all tool names are unique."
```
ahmetgunduz, line 140: Can we make this check for all the list and then do assert for the non unique ones for the whole list of tools? Current way will do check on by one and if I have multiple uplicated pairs I will have to to multiple iteration of error-fix-error-fix etc. 

Following can be better. 

1. Create tool list of names
2. Check duplications
3. If there exists any, assert with all of them in error message

aixplain/factories/agent_factory/__init__.py
```diff
@@ -179,19 +180,23 @@ def create_model_tool(
...
-    def create_custom_python_code_tool(cls, code: Union[Text, Callable], description: Text = "") -> CustomPythonCodeTool:
+    def create_custom_python_code_tool(
+        cls, code: Union[Text, Callable], name: Text, description: Text = ""
+    ) -> CustomPythonCodeTool:
         """Create a new custom python code tool."""
-        return CustomPythonCodeTool(description=description, code=code)
+        return CustomPythonCodeTool(name=name, description=description, code=code)
 
     @classmethod
     def create_sql_tool(
```
ahmetgunduz, line 202: why not have a name for sql as well.

aixplain/factories/agent_factory/__init__.py
ahmetgunduz, line 1: I would put name as the first parameter for all the functions.

aixplain/v2/agent.py
ahmetgunduz, line 1: same comments as for AgentFactory ones

aixplain/modules/agent/tool/model_tool.py
thiago-aixplain: it will be the model name, since in this case `model` won't be None

aixplain/modules/agent/tool/custom_python_code_tool.py
thiago-aixplain: I am checking this in the validation.

aixplain/modules/agent/tool/pipeline_tool.py
thiago-aixplain: changed for:

```
self.name = self.name if self.name else set_tool_name(self.function, self.supplier, self.model)
```

aixplain/modules/agent/tool/python_interpreter_tool.py
thiago-aixplain: I am calling it `"Python Interpreter"` now.

aixplain/modules/agent/__init__.py
thiago-aixplain: Done

aixplain/factories/agent_factory/__init__.py
thiago-aixplain: I am afraid to break old versions

aixplain/factories/agent_factory/__init__.py
thiago-aixplain: Done

aixplain/factories/agent_factory/utils.py
```diff
@@ -60,14 +60,20 @@ def build_tool(tool: Dict):
         else:
             tool = PythonInterpreterTool()
     elif tool["type"] == "sql":
+        name = tool.get("name", "")
```
ahmetgunduz, line 63: Lets have default name for any tool not empty string but a proper name. For example, here it can be  name = tool.get("name", "SQLTool")

ahmetgunduz (approved):

aixplain/modules/agent/tool/custom_python_code_tool.py
ahmetgunduz: Still the default name is empty string  though, am I missing sth?

PR #506, opened 2025-04-28
ENG-2028: model streaming

thiago-aixplain:


aixplain/modules/model/__init__.py
```diff
@@ -232,14 +237,28 @@ def poll(self, poll_url: Text, name: Text = "model_process") -> ModelResponse:
...
             )
 
+    def run_stream(
+        self,
+        data: Union[Text, Dict],
+        parameters: Optional[Dict] = None,
+    ) -> ModelResponseStreamer:
+        assert self.supports_streaming, f"Model '{self.name} ({self.id})' does not support streaming"
+        payload = build_payload(data=data, parameters=parameters, stream=True)
+        url = f"{self.url}/{self.id}".replace("api/v1/execute", "api/v2/execute")
```
ahmetgunduz, line 247: Can we have a warning around this? 
I suppose only v2 supports streaming that is why?

aixplain/modules/model/model_response_streamer.py
```diff
@@ -0,0 +1,28 @@
+import json
+from typing import Iterator
+
+from aixplain.modules.model.response import ModelResponse, ResponseStatus
+
+
+class ModelResponseStreamer:
+    def __init__(self, iterator: Iterator):
```
ahmetgunduz, line 8: Can we have all the other Model Response params also set for this? 
data=response.pop("data", ""),
details=response.pop("details", {}),
completed=response.pop("completed", False),
error_message=response.pop("error_message", ""),
used_credits=response.pop("usedCredits", 0),
run_time=response.pop("runTime", 0),
usage=response.pop("usage", None),
error_code=response.get("error_code", None),

ahmetgunduz (approved):

aixplain/modules/model/model_response_streamer.py
thiago-aixplain: They are by default.

PR #524, opened 2025-05-12
ENG-2003 : Add LLM's to Agents as Object

ahmetgunduz:


aixplain/factories/agent_factory/__init__.py
```diff
@@ -53,7 +55,8 @@ def create(
         name: Text,
         description: Text,
         instructions: Optional[Text] = None,
-        llm_id: Text = "669a63646eb56306647e1091",
+        llm: Optional[LLM] = None,
```
thiago-aixplain, line 58: maybe considering llm_id here as well? `Union[LLM, Text]`

aixplain/factories/agent_factory/utils.py
```diff
@@ -100,6 +100,39 @@ def build_agent(payload: Dict, tools: List[Tool] = None, api_key: Text = config.
                 )
                 continue
 
+    # Get LLM from tools if present
```
thiago-aixplain, line 103: Encapsulate the logic to instantiate the LLM in a single function.

ahmetgunduz:
here is the colab: https://colab.research.google.com/drive/1hmM3r79nCtosSdU9vE-ad68F0Ld_z8dt?usp=sharing#scrollTo=8vjEqugb_-yd

PR #534, opened 2025-05-20
ENG-2161: Update Model Params for Embeddings

basitanees:
Add attributes to model card additional_info key:

Example Usage:
- Getting embedding size of embedding models from the model card
- Getting embedding model, size in IndexModel

### DIFF
--- a/aixplain/modules/model/index_model.py
+++ b/aixplain/modules/model/index_model.py
@@ -83,6 +83,24 @@ class IndexModel(Model):
         self.url = config.MODELS_RUN_URL
         self.backend_url = config.BACKEND_URL
         self.embedding_model = embedding_model
+        if embedding_model:
+            try:
+                from aixplain.factories import ModelFactory
+
+                model = ModelFactory.get(embedding_model)
+                self.embedding_size = model.additional_info["embedding_size"]
+            except Exception as e:
+                import warnings
+
+                warnings.warn(f"Failed to get embedding size for embedding model {embedding_model}: {e}")
+                self.embedding_size = None
+
+    def to_dict(self) -> Dict:
+        data = super().to_dict()
+        data["embedding_model"] = self.embedding_model
+        data["embedding_size"] = self.embedding_size
+        data["collection_type"] = self.version.split("-", 1)[0]
+        return data
 
     def search(self, query: str, top_k: int = 10, filters: List[IndexFilter] = []) -> ModelResponse:
         """Search for documents in the index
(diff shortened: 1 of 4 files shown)


aixplain/modules/model/index_model.py
```diff
@@ -83,6 +83,24 @@ def __init__(
         self.url = config.MODELS_RUN_URL
         self.backend_url = config.BACKEND_URL
         self.embedding_model = embedding_model
+        if embedding_model:
+            try:
+                from aixplain.factories import ModelFactory
+
+                model = ModelFactory.get(embedding_model)
+                self.embedding_size = model.additional_info["embedding_size"]
```
thiago-aixplain, line 91: self.embedding_size = model.additional_info.get("embedding_size")

aixplain/modules/model/index_model.py
```diff
@@ -83,6 +83,24 @@ def __init__(
...
+                import warnings
+
+                warnings.warn(f"Failed to get embedding size for embedding model {embedding_model}: {e}")
+                self.embedding_size = None
+
+    def to_dict(self) -> Dict:
+        data = super().to_dict()
+        data["embedding_model"] = self.embedding_model
+        data["embedding_size"] = self.embedding_size
+        data["collection_type"] = self.version.split("-", 1)[0]
```
thiago-aixplain, line 102: is there any case that version is `None`? In case not, this is fine. Otherwise, it will crash.

tests/functional/model/run_model_test.py
```diff
@@ -80,6 +81,21 @@ def test_run_async():
...
+    "embedding_model,supplier_params",
+    [pytest.param(EmbeddingModel.OPENAI_ADA002, AirParams, id="AIR - OpenAI Ada 002")],
+)
+def test_index_model_with_embedding(embedding_model, supplier_params):
+    from uuid import uuid4
+    from aixplain.factories import IndexFactory
+
+    params = supplier_params(name=str(uuid4()), description=str(uuid4()), embedding_model=embedding_model)
+    index_model = IndexFactory.create(params=params)
+    assert index_model.embedding_model == embedding_model
```
thiago-aixplain, line 94: what will happen if I add `index_model = ModelFactory.get(index_model.id)` before the assertions?

basitanees: @thiago-aixplain It should work the same way. IndexFactory.create also actually also returns a ModelFactory.get in the end

aixplain/modules/model/index_model.py
basitanees: version shouldnt be none ever as it points to collection name and environment name in the format:
{index_type}-{env}-{team}-{collection_name}

aixplain/modules/model/index_model.py
basitanees: @thiago-aixplain In that case, let me update this as try catch doesnt need to check for embedding_size key retrieval. It can only check if ModelFactory.get doesnt fail

thiago-aixplain (approved):

PR #550, opened 2025-06-13
ENG-2271: Add code interpreter model ID

lucas-aixplain:


aixplain/enums/code_interpeter.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
```
ahmetgunduz, line 1: Can you add this into enums/__init__.py?

ahmetgunduz (approved):

PR #551, opened 2025-06-16
Eng 2281 Add file parsing for Index

basitanees:


basitanees:
docling model not working and tests failing therefore

PR #603, opened 2025-07-21
session id agent init changes

xainaz:


aixplain/modules/agent/__init__.py
```diff
@@ -281,6 +298,7 @@ def run_async(
         max_iterations: int = 10,
         output_format: OutputFormat = OutputFormat.TEXT,
         expected_output: Optional[Union[BaseModel, Text, dict]] = None,
+        allow_history_and_session_id: Optional[bool] = False
```
ahmetgunduz, line 301: This  should always be False, isnt it. why we get this as a parameter that user can  set it to true?

ahmetgunduz (changes requested):

aixplain/modules/agent/__init__.py
xainaz: We added this parameter because the issue Tim pointed out. 
If we want the chat history to be correctly formatted, like this:
chat_history = [
    HumanMessage(content="What is a black hole?"),
    AIMessage(content="A black hole is..."),
    HumanMessage(content="How do black holes form?"),
    AIMessage(content="They typically form..."),
]

Then we need to be able to send the request to make a session id with provided history to engine with both history and session id, which we disabled (throws an error if we do so)
I suggested we either connect the generate_session_id on SDK to the one on engine which directly adds history to the session id without throwing an error, but this would include lots of work on backend. Or we could add this parameter to allow generate_session_id to send both to run. You suggested the second option until we find a better solution. It doesn't make a lot of sense, but it's the only solution that wouldn't involve work on backend I think.

aixplain/modules/team_agent/__init__.py
```diff
@@ -335,6 +340,29 @@ def delete(self) -> None:
...
+                # Ensure it's actually a dictionary and not a Mock or other object
+                if isinstance(agent_dict, dict) and hasattr(agent_dict, "items"):
+                    try:
+                        # Add all fields except 'id' to avoid duplication with 'assetId'
+                        additional_data = {k: v for k, v in agent_dict.items() if k not in ["id"]}
+                        base_dict.update(additional_data)
+                    except (TypeError, AttributeError):
+                        # If items() doesn't work or iteration fails, skip the additional data
+                        pass
+        except Exception:
```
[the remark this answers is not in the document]
xainaz: I think this came from resolving a merge conflict. I didn't add it but it's on dev right now.

ahmetgunduz (approved):

PR #863, opened 2026-03-11
Add backend-managed session CRUD, messages, reactions, and file uploads

aix-ahmet:
Introduce Session and SessionMessage resources with full lifecycle support
  (create, get, list, update, delete), message operations (add, list, get,
  delete), reactions (like/dislike/clear), and file upload attachments.
  Deprecate client-side generate_session_id() in favor of create_session().

  Includes 87 unit tests and functional test suite.


kadirpekel (approved):

aix-ahmet (commented):
## Blocking / Must-fix

### 1. `generate_session_id()` is **removed**, not deprecated — breaking API change
The PR body says *"Deprecate client-side `generate_session_id()` in favor of
`create_session()`."* The diff actually **deletes** the v2 method outright and
replaces it with `create_session`. Any existing caller of
`agent.generate_session_id()` on v2 will get `AttributeError` with no migration
window. (v1's `generate_session_id` is untouched — separate code path.)

Additional behavioral break: the old method auto-saved the agent
(`self.save(as_draft=True)`); `create_session` now **raises** `ValueError` if the
agent is unsaved. Callers relying on the implicit save will break.

**Action:**  this is a hard break , say so explicitly in the PR body + CHANGELOG and bump the version accordingly. Don't ship a hard removal labeled "deprecate."

---

## Should-fix

### 2. Reused sessions silently ignore per-run execution params
In `_run_via_session`, when a `session_id` is supplied the code does
`session = self.context.Session.get(session_id)` and then `add_message(...)`. The
`execution_params` / `criteria` / `evolve` / `run_response_generation` kwargs the
caller passed to `run()` are **only** applied when a session is *created* (the
`else` branch). On a reused session they're silently dropped — the run executes
with whatever `executionConfig` the session was created with.

**Action:** Update the session's `executionConfig` before posting
the message if per-run overrides should take effect. At minimum, warn when override
kwargs are passed alongside an existing `session_id`.

elsheikhams99 (approved):

PR #934, opened 2026-05-22
error codes changes

xainaz:


aixplain/v1/modules/agent/__init__.py
```diff
@@ -677,6 +677,7 @@ def run(
                 progress_verbosity=progress_verbosity,
             )
             result_data = result.get("data") or {}
+            error_codes = result_data.get("errorCodes", [])
```
aix-ahmet: Two concerns (applies to `team_agent/__init__.py` too):

1. The engine emits snake_case `error_codes` on its response, and platform-backend#2002 only touches logging + GraphQL trace entities — I couldn't find the layer that camelizes and nests `errorCodes` under `data` for the poll payload. Can we run one real `agent.run()` against dev and confirm this isn't always `[]`? Cheap defensive fix regardless: `result_data.get("errorCodes") or result_data.get("error_codes", [])` — the backend logger already reads both casings, the SDK should too.
2. This only covers the sync `run()` path — users doing `run_async()` + `poll()` never get `error_codes`. Better to populate it where the poll response is parsed so both paths benefit.

aix-ahmet (commented):
Reviewed together with aixplain/agentification#718 and aixplain/platform-backend#2002 (PROD-2328). The SDK changes themselves are clean and the tests are good. Two things before merge (see inline): the `errorCodes` casing/nesting needs to be verified end-to-end against a dev deployment, and the async (`run_async` + `poll`) path doesn't surface `error_codes` at all.

Also the PR description is empty — please link PROD-2328 and document the new field.

aix-ahmet (approved):

PR #954, opened 2026-06-10
feat(v2): send per-run budget to agents via execution_params (#953)

### LINKED ISSUE #953: [v2][budget] SDK: send per-run budget to agents via execution_params
# [v2][budget] SDK: send per-run budget to agents via `execution_params`

## Goal
Let SDK users attach a **budget** (cost / duration / iteration caps) to a single agent
run. The budget feature already exists and is enforced in the `aixplain-agents`
execution service (`BudgetConfig` + `BudgetController`). This task only makes the
**Python SDK produce** the budget payload. It is **run-time only** — not persisted on
the agent.

## Budget contract — SHARED, DO NOT DEVIATE
This exact shape is agreed across three repos (SDK ← this task, `platform-backend`,
`aixplain-agents`). If any key differs, budget is silently ignored downstream.

Budget rides inside `executionParams` as a nested `budget` object:

```jsonc
"executionParams": {
  "maxTokens": 2048,
  "maxIterations": 5,
  "budget": {
    "maxCost": 1.0,            // float, USD spending cap for the run
    "maxDurationSeconds": 300, // float, wall-clock seconds
    "maxIterations": 50        // int, node-execution cap (budget guardrail)
  }
}
```

- All wire keys are **camelCase**. `budget` and every field inside it are **optional**.
- Omitting `budget` ⇒ no budget enforcement (current behavior — must stay backward compatible).
- `warnAtPercent` is intentionally **NOT** sent; the backend `BudgetConfig` default `0.8` applies.
- Python user-facing API is snake_case (`max_cost`, `max_duration_seconds`, `max_iterations`)
  and serializes to the camelCase keys above.

## Scope (this repo only)
`aixplain/v2/agent.py`

1. Add a small serializable `Budget` type (dataclass, snake_case fields
   `max_cost: Optional[float]`, `max_duration_seconds: Optional[float]`,
   `max_iterations: Optional[int]`) that serializes to the camelCase keys
   `maxCost` / `maxDurationSeconds` / `maxIterations`, dropping `None` fields.
   Follow the existing `@dataclass_json` + `metadata=config(field_name=...)` pattern
   used elsewhere in this file.
2. Add `budget: NotRequired[Optional[Union[Dict, Budget]]]` to `AgentRunParams`
   (run-time params, ~line 218). Accept either a `Budget` instance or a plain dict.
3. In `build_run_payload` (~line 1229): pop `budget` from kwargs, normalize it to the
   nested camelCase dict, and set `execution_params["budget"]` only when provided.
   Do **not** add a default — absence must produce no `budget` key.
   Reuse / extend `_EXEC_PARAMS_MAP` if a snake→camel mapping for `budget` is needed.
4. Keep `max_cost`/`max_duration_seconds`/`max_iterations` snake→camel conversion local
   to the `Budget` serialization so user-supplied dicts in either case work.

## Out of scope
- No creation-time / persisted budget field on `Agent`.
- No platform-backend or agents-service changes (separate tasks).
- Do not expose `warnAtPercent`.

## Acceptance criteria
- `agent.run("q", budget=Budget(max_cost=1.0, max_duration_seconds=300, max_iterations=50))`
  produces a run payload whose `executionParams.budget == {"maxCost":1.0,"maxDurationSeconds":300,"maxIterations":50}`.
- Passing `budget` as a plain dict (snake_case or camelCase) yields the same camelCase payload.
- Partial budgets work: `Budget(max_cost=2.0)` ⇒ `executionParams.budget == {"maxCost":2.0}` (no null keys).
- No `budget` passed ⇒ no `budget` key in `executionParams` (byte-for-byte unchanged from today).
- New unit test file `tests/unit/v2/test_agent_budget.py` covering the above.

## Verification
`python -m pytest tests/unit`

aix-ahmet:
Closes #953

## Summary

Adds the ability to send a **per-run budget** to agents. The budget is run-time only (never persisted on the `Agent`) and is serialized into `executionParams.budget` on the wire.

## Changes

- **New `Budget` dataclass** with optional fields `max_cost` / `max_duration_seconds` / `max_iterations`, serializing to the agreed camelCase wire keys `maxCost` / `maxDurationSeconds` / `maxIterations`. `None` fields are dropped from `to_dict()`.
- **`AgentRunParams`** accepts an optional `budget` — either a `Budget` instance or a plain dict (snake_case or camelCase).
- **`build_run_payload`** normalizes the budget (`_normalize_budget`) and sets `executionParams.budget` only when non-empty; absence leaves the payload unchanged. Unknown keys pass through for forward compatibility.
- **`Budget` exported** from `aixplain.v2`.

## Usage

```python
from aixplain.v2 import Budget

agent.run("...", budget=Budget(max_cost=0.5, max_iterations=10))
# or a plain dict
agent.run("...", budget={"max_cost": 0.5})
```

## Tests

New `tests/unit/v2/test_agent_budget.py` covering serialization (camelCase mapping, None-dropping) and payload behavior (presence only when provided, dict/instance/snake/camel inputs). All 17 tests pass.


aix-ahmet:
## Review findings

Adversarially-verified review. The core reshape checks out: budget IS persisted on explicit `save()` and sent on every run, and the `is_modified` exclusion correctly implements "budget mutations must not block runs". Four confirmed issues and two judgment calls:

**Confirmed:**

1. **`run(budget=...)` is silently swallowed** — `build_run_payload` does `kwargs.pop("budget", None)` with no warning, while the agent's own (possibly empty) budget is sent. That's the most natural call for a per-run cap and it no-ops — the exact failure budgets exist to prevent. One `raise`/`warn` pointing at `agent.budget.max_cost = ...` would close it.
2. **`_coerce_budget` erases unknown budget keys** — every dict is coerced to the 3-field `Budget`, so a backend budget carrying any future field is dropped on a get→save round-trip. `_normalize_budget`'s "unknown keys pass through" comment only applies to raw dicts, which never survive coercion.
3. **`ExecutionConfig.budget` is write-only** — `to_api_dict` nests it into `executionParams.budget`, but reload never lifts it back, so `session.execution_config.budget` reads `None` on every fetched session while the value hides inside `execution_params["budget"]`.
4. **`agent.max_iterations` reads `None` after fetch** — `_fold_legacy_max_iterations` moves a persisted `maxIterations` into `budget` and never writes the attribute back (the PR's own test codifies `agent.max_iterations is None`). Legacy code reading the attribute breaks silently.

**Judgment calls worth a deliberate decision:**

5. `_apply_run_overrides_to_session` merges without `budget` in the base dict — a reused session run with an `execution_params` override re-saves executionConfig without the nested budget cap (criteria-only overrides happen to be safe because the old `execution_params` dict is carried whole).
6. The client-side default `executionParams.maxIterations=5` was removed entirely; agents with an empty budget now send no cap and defer to the backend default. If that default isn't 5, run behavior changes silently for every existing agent — worth stating in the PR description if intended.

**Cleanup:** the wrapped-`from_dict` shim is now copy-pasted three times (Agent, ExecutionConfig, Session) and the deprecated max-iterations fold block is duplicated between `build_run_payload` and `to_api_dict` — both are one shared helper away from drift-proof.

A follow-up branch with fixes + tests for items 1–3 and 5 is being prepared (`review-fixes/issue-953-budget`).

PR #962, opened 2026-06-16
feat: add v2 Skill asset (knowledge file bundle)

aix-ahmet:
## Summary

Adds `aix.Skill`, a new **v2 managed asset** backed by the `/sdk/skill` backend endpoints. A Skill is a curated bundle of files and folders (a "knowledge bundle") that can be attached to an Agent to give it reference material. Designed against the v2 design canon — it gets/searches/saves/deletes like every other asset and attaches to agents the same way tools/sub-agents do.

## Design highlights

- **`Skill`** composes `BaseResource + Search + Get + Delete` mixins. It is **not runnable** (no `/run` endpoint exists), so it deliberately omits `RunnableResourceMixin`.
- **Lifecycle** is the standard `instantiate → mutate → save()`: `aix.Skill(...).save()` (create/update auto-detected), `.get(path_or_id)`, `.search(query=, tags=, suppliers=, saved=)` → `Page`, `.delete()`, `.clone()`.
- **File tree** is modeled as composition: `SkillNode` + the `aix.Skill.File(...)` / `aix.Skill.Folder(...)` constructors (mirrors `Agent.Task`). Files accept a remote `url=` or a local `path=` (auto-uploaded via `FileUploader`).
- **`save(save_subcomponents=True)` (BETA)** reconciles `children` against the last-synced state and issues the backend node calls (create folder/file, update file url, delete node) — the per-node routes stay hidden from developers.
- Convenience: `skill.download(to=...)`, `skill.get_node("docs/faq.pdf")`, `skill.refresh()`.
- **`Agent.skills`**: `List[Union[str, Skill]]`, passed exactly like `tools`/`agents`. Resolves to ids on serialization, participates in `save(save_subcomponents=True)`, and raises the standard "must be saved first" validation for unsaved skills.
- Wired into `core.py` (`aix.Skill`) and v2 `__init__` exports (`Skill`, `SkillNode`).

## Usage

\`\`\`python
skill = aix.Skill(name="Support Playbook", tags=["support"], privacy=aix.Privacy.PRIVATE)
skill.save()

docs = aix.Skill.Folder(name="docs")
skill.children.append(docs)
skill.children.append(aix.Skill.File(name="faq.pdf", path="./faq.pdf", parent=docs))
skill.save(save_subcomponents=True)

agent = aix.Agent(name="Support Agent", skills=[skill, "60f...skill_id"])
agent.save()
\`\`\`

## Notes / follow-ups

- **`get(path)`** is a thin pass-through to the backend `{id}` slot (same as `Model`). It works with ids today; path resolution depends on the backend resolving a `vendor/name`-style path in that slot (tracked separately).
- `download(...)` introduces a new verb not yet in the design canon — proposing it be added so future file-bearing assets reuse the signature.

## Testing

- Full v2 unit suite passes (`719 passed, 4 skipped`).
- Pre-commit (ruff, ruff-format, pytest) green.
- Smoke-tested client wiring, save-payload exclusion of `children`, node constructors, `get_node`, and `Agent(skills=[...])` id resolution + validation.


aix-ahmet:
## Review findings

Adversarially-verified review. One item escalates the issue already tracked with @Hadi — it's worse than "retrieved agent lacks skills":

**Agent-side skills handling:**

1. **`Agent.get(id).save()` strips the agent's skills from the update payload** — `_original_skills` is only populated in `__post_init__` from `self.skills`; a fetched agent (backend GET carries no skills) has it empty, so `build_save_payload` hits the `payload.pop("skills", None)` branch and the update is sent without them. Depending on backend replace semantics this is silent data loss on any fetch→edit→save cycle, not just a missing field on read. Worth folding into the open backend discussion.
2. **`agent.skills = [my_skill]` after construction is silently ignored** — the payload iterates frozen `_original_skills`, unlike tools which are read live from `self.tools`. Assignment works for tools and no-ops for skills.
3. **`self.skills` stays `[None]` after saving an unsaved skill** — ids are flattened in `__post_init__` before subcomponents have ids; `_save_subcomponents` assigns ids on the `_original_skills` objects but `self.skills` is never refreshed (and `_saved_state` baselines `[None]`).

**skill.py hardening:**

4. `_load_from_path` uses `abspath` without `expanduser` — `Skill(file_path="~/skills/x")` fails with "Skill path not found".
5. Frontmatter parsing is unguarded — malformed YAML propagates `yaml.YAMLError` out of the constructor, and scalar/list frontmatter (truthy, so `or {}` doesn't catch it) raises `AttributeError` on `meta.get`.
6. The single-file branch accepts **any** existing file (no `.md` check) and registers it as `SKILL.md` with the original file's inferred content-type — `Skill(file_path="notes.txt")` passes silently, contradicting the documented contract.
7. `download()` default `./{self.name}.zip` breaks for path-style names (`ws/skill` → FileNotFoundError on the missing dir) and writes `./None.zip` when a fetched skill has no name.

**Cleanup:** `_upload()` re-implements the `FileUploader` wiring `file.py`/`upload_utils` already provide, and builds a new uploader per file inside the walk; the file-vs-folder upload branches share ~90% logic (materializing a lone .md into a temp bundle dir would collapse them); `download()` buffers the whole zip in memory rather than streaming.

A follow-up branch with fixes + tests for items 4–7 is being prepared (`review-fixes/v2-skill-asset`); items 1–3 need a design decision (live vs frozen skills serialization) so they're flagged here instead.

PR #965, opened 2026-06-19
added warning for saving agent with integration

xainaz:


aixplain/v2/agent.py
```diff
@@ -830,8 +830,45 @@ def save(self, *args: Any, **kwargs: Any) -> "Agent":
         # Validate that all dependencies are saved before proceeding
         self._validate_dependencies()
 
+        unconnected_integration_names = self._get_unconnected_integration_names()
+
         # Call the parent save method
-        return super().save(*args, **kwargs)
+        saved_agent = super().save(*args, **kwargs)
+        self._warn_for_unconnected_integrations(unconnected_integration_names)
```
aix-ahmet, line 838: **Minor: the pre-save compute + pass-through only pays off if `super().save()` mutates `self.tools`**

`unconnected_integration_names` is computed here (before save) and threaded into `_warn_for_unconnected_integrations`, which otherwise re-derives it. That split is only necessary if `super().save()` replaces the in-memory `Integration` objects (e.g. swaps them for connected tools/ids), so the names must be captured beforehand. If save leaves `self.tools` untouched, this could collapse to a single post-save `self._warn_for_unconnected_integrations()` (no arg).

Either way, a one-line comment on *why* it's captured pre-save would save the next reader the trace. (Also: `_get_unconnected_integration_names` flags every `Integration` instance and the warning says "is not connected" — fine if an `Integration` used directly as a tool is unconnected by definition, just calling it out since the wording asserts a state that's never checked.)

PR #973, opened 2026-06-24
Eng 3203 attachments without session

elsheikhams99:


aixplain/v2/agent.py
```diff
@@ -245,6 +246,11 @@ class AgentRunParams(BaseRunParams):
         evolve: Evolution parameters
         inspectors: Inspector configurations
         run_response_generation: Whether to run response generation. Defaults to False.
+        attachments: Multimodal attachments for the turn (only with ``via_session=True``).
```
aix-ahmet: why it says session only?

aix-ahmet (changes requested):

PR #981, opened 2026-07-01
feat(v2): object-based tool parameter overrides

elsheikhams99:


aix-ahmet:
## Review findings

Adversarially-verified review of this PR (each item below was confirmed against the diff and branch code). The object-mutation design is a clear improvement over the #967 dict shape; the issues are all in the seams where the old surface was removed or where the run-time serializer diverges from the save path.

**Correctness (most severe first):**

1. **Stray `run(tools=[...])` leaks raw into the payload** — `build_run_payload` no longer pops `tools`, so the generic snake→camel kwargs loop forwards the list verbatim (dict-shaped `parameters`, never normalized), and `_build_tool_overrides()` only overwrites `payload["tools"]` when non-empty. A caller on the old API sends a malformed payload with no error. Suggest rejecting `tools=` explicitly.
2. **`run(via_session=True, tools=[...])` is silently dropped** — `_run_via_session` no longer reads it and `tools` is not in `_LEGACY_ONLY_RUN_KWARGS`, so the offending-kwargs guard never fires. This re-introduces the PROD-2481 "silently dropped tools" class. Adding `"tools"` to the legacy tuple fixes this and gives item 1 a clean error on the session path too.
3. **`_normalize_tool_dict_for_api` lost its dict-`parameters` branch** — an unbound `Agent(tools=[{"id":"t","type":"tool","parameters":{"a":1}}])` (no context → no hydration) now emits the raw `{"a":1}` dict in the save payload instead of the `[{name,value}]` list.
4. **Hydration destroys legacy `{id, parameters:{key:value}}` values on a bound agent** — `_build_tool_tool` iterates the flat dict's keys, each fails `isinstance(action_def, dict)`, and the Tool comes out with empty actions; the caller's values silently vanish from both save and run.
5. **`_current_tool_parameters` flattens all actions into one `{name: value}` map** — two materialized actions sharing an input name collide (last write wins), and the emitted run override carries no action code, unlike the save path's per-action `[{code, name, inputs}]` shape. Multi-action tools lose information at run time.
6. **Fetched agents re-send persisted params as an "override" on every run** — hydration seeds `Input.value` from the snapshot, and the serializer can't tell seeded from user-set, so every `run()` carries a redundant (potentially stale) override, contradicting the "no-override runs stay lightweight" docstring.

**Smaller/cleanup:**

7. `_original_tools` is assigned in `__post_init__` and twice in `save()` but never read — dead state.
8. `_current_tool_parameters` reaches into private `_actions`/`_inputs` internals of `actions.py`; a rename there silently breaks run overrides. An offline accessor on Tool/Model shared with `get_parameters()` would be sturdier.
9. The tools null-out/restore dance now exists in both `build_save_payload` and `_get_serializable_state`; any other `to_dict()` caller still crashes on object tools. A field-level encoder/exclude (the pattern `_role_field` already uses) would fix all callers at once.

A follow-up branch with fixes + tests for items 1–4 and 7 is being prepared (`review-fixes/ENG-3228-tool-params`).

PR #984, opened 2026-07-09
feat(v2): manage sessions via aix.Session; polymorphic agent.run(session=)

elsheikhams99:
1. Manage sessions through aix.Session alone — removed agent.create_session(), agent.list_sessions(), agent.generate_session_id(), and the internal _resolve_execution_config() helper. Added an agent= init arg to [Session](vscode-webview://0v0m0cuimos6sv6di2jcmtv8mfn5vlksl9rbir21dbi0ueb2327d/aixplain/v2/session.py) (accepts an Agent or an id string), so aix.Session(agent=agent, name=…) is the single create path.

2. Session.search(agent=…) -> Page — replaced the bespoke Session.list(). Session now inherits SearchResourceMixin and overrides [search()](vscode-webview://0v0m0cuimos6sv6di2jcmtv8mfn5vlksl9rbir21dbi0ueb2327d/aixplain/v2/session.py) to hit the real GET /v1/sessions endpoint and wrap results in a Page (tolerates a future paginated-dict envelope too).

3. Date-range + memory filters — search(created_after=…, created_before=…, memory_enabled=…) are wired into the query params. (SDK-side only; backend filter delivery is a follow-up per the story, noted in the docstring.)

4. agent.run(query, session=…) — polymorphic (Session object or id string) via the new _resolve_session/_run_with_session. Dropped via_session and the id-only session_id= from AgentRunParams and the snake→camel map; run_async rejects session= as before.

6. Removed agent.generate_session_id() — gone.

---
IMPORTANT:
- Item 5 (one run endpoint): session= runs go through POST /v1/sessions/{id}/messages; sessionless runs still use the direct /v2/agents/{id}/run. Retiring the direct endpoint needs Ibrahim / backend consolidation.
- Item 7 (ownership): built agent-bound now (matches the backend that welds a session to one agent_id). Agent-agnostic threads / handoff are deferred until the backend can drop the binding — the agent= arg gives us a clean seam to relax later.

### DIFF
--- a/aixplain/v2/agent.py
+++ b/aixplain/v2/agent.py
@@ -231,11 +231,13 @@ class AgentRunParams(BaseRunParams):
     """Parameters for running an agent.
 
     Attributes:
-        session_id: Session ID for conversation continuity
+        session: Conversation thread to run within. A
+            :class:`~aixplain.v2.session.Session` instance or a session id
+            string. Omit for a one-shot, stateless run. Replaces the removed
+            ``via_session`` flag and id-only ``session_id``.
         query: The query to run
         variables: Variables to replace {{variable}} placeholders in instructions and description.
             The backend performs the actual substitution.
-        allow_history_and_session_id: Allow both history and session ID
         tasks: List of tasks for the agent
         prompt: Custom prompt override
         history: Conversation history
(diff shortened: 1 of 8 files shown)


aixplain/v2/agent.py
```diff
@@ -259,10 +261,9 @@ class AgentRunParams(BaseRunParams):
         progress_truncate: Whether to truncate long text in progress display
     """
 
-    session_id: NotRequired[Optional[Text]]
```
aix-ahmet, line 262: should we depracate it before removing? Need to be backward compatiable

aixplain/v2/agent.py
```diff
@@ -272,7 +273,6 @@ class AgentRunParams(BaseRunParams):
     identifier: NotRequired[Optional[Text]]
     inspectors: NotRequired[Optional[List[Dict]]]
     run_response_generation: NotRequired[Optional[bool]]
-    via_session: NotRequired[Optional[bool]]
```
aix-ahmet, line 275: good

aix-ahmet (approved):

PR #1040, opened 2026-08-12
BUG-937: v2 — validate poll-URL host and drop x-api-key on cross-host redirect (SEV1)

aix-ahmet:
## Problem

In v2, poll URLs are taken from **response bodies**. `AixplainClient.request_raw` and `request_stream` both accepted an absolute URL verbatim, while the API key lived permanently on the shared `requests.Session` headers. Two consequences:

1. **Body-supplied host.** Any JSON field that named a host (`url`, `data`, `poll_url`, …) would cause the SDK to send `TEAM_API_KEY` to that host — the session headers ride along with whatever absolute URL is handed to the client.
2. **Cross-host redirect.** A single `302` from a trusted host leaked the key too: `requests` strips only `Authorization` on a host change. The custom `x-api-key` / `x-aixplain-key` headers are copied verbatim to the redirect target.

Additionally, `Model._is_poll_url` matched host *and* path in one regex, so a poll-shaped URL on a foreign host silently fell through to being treated as **output media data** — hiding an attack rather than failing loudly.

## Fix (v2 only)

**Credentials off the session.** `AixplainClient` keeps only `Content-Type` on the session; auth headers are attached per request, *after* the target URL has been validated.

**Host allowlist, enforced before the socket opens.**
- `ensure_trusted_url(path)` resolves the path and raises `UntrustedURLError` unless the result is a trusted origin. Validation is on the **resolved** URL — a relative input like `//evil.example.com/x` only becomes absolute after `urljoin`, so checking the caller's raw string would be bypassable.
- Trusted origins are `(scheme, host, port)` triples built from `base_url`, `model_url`, `pipeline_url`, the aiXplain defaults, and the `AIXPLAIN_TRUSTED_HOSTS` env var (for on-prem / regional deployments whose poll hosts aren't derivable from the configured endpoints). `AixplainClient(trusted_urls=[...])` is also available programmatically.
- `normalize_origin` parses via `hostname`/`port` rather than `netloc`, so userinfo cannot disguise the target (`https://platform-api.aixplain.com@evil.example.com/x` → **rejected**) and `:443` compares equal to the implicit default. Scheme is part of the origin, so plain `http` to an aiXplain host is untrusted unless an operator explicitly configured an `http` endpoint.

**Redirect stripping.** `_AixplainSession.rebuild_auth` pops both credential headers when the redirect target is outside the trusted set. Redirects stay **enabled** — a legitimate `302` to a presigned S3 result URL still resolves, it just arrives unauthenticated (presigned URLs carry their own signature). The allowlist is a class-level `frozenset()` default and is listed in `__attrs__`, so a session restored via deepcopy/pickle fails **closed**.

**Polling path.**
- `RunnableResourceMixin.poll` validates the poll URL up front, so the error stays precise instead of being rewrapped as `"Polling failed: …"`.
- `UntrustedURLError` is re-raised from the polling loop; without that, the broad `except Exception: continue` would have turned a refused credential leak into a silent poll-until-timeout.
- `Model._is_poll_url` now matches on **path shape only** — routing, never authorization. A poll-shaped foreign URL matches here and is then refused by the trust guard, which is the desired loud failure.

**Logging.** Now that credentials travel in `kwargs`, the debug log line redacts them (`***`) instead of printing the key.

## API surface

- New `aixplain.v2.UntrustedURLError` (subclass of `AixplainV2Error`, deliberately **not** an `APIError` — no request is made, so there's no status code).
- New optional `AixplainClient(trusted_urls=[...])` kwarg and `AIXPLAIN_TRUSTED_HOSTS` env var.
- No breaking change for normal use: all legitimately-configured endpoints (backend, model, pipeline) are trusted automatically.

## Tests

New `tests/unit/v2/test_url_guard.py` (378 lines) and `tests/unit/v2/test_poll_url_guard.py` (257 lines) cover origin normalization, env-var configuration, redirect header stripping, userinfo / explicit-port / `:0` / malformed-port edge cases, the poll-path guard, and the loop's re-raise behavior. `test_client.py` and `test_core.py` updated for the per-request header model.

**Verification:** `pytest tests/unit/v2/` → **1190 passed, 4 skipped**. Full pre-commit suite (ruff, ruff format, pytest-check) passes on the commit.

### DIFF
--- a/aixplain/v2/__init__.py
+++ b/aixplain/v2/__init__.py
@@ -61,6 +61,7 @@ from .exceptions import (
     ValidationError,
     TimeoutError,
     FileUploadError,
+    UntrustedURLError,
 )
 from .enums import (
     AuthenticationScheme,
@@ -161,6 +162,7 @@ __all__ = [
     "ValidationError",
     "TimeoutError",
     "FileUploadError",
+    "UntrustedURLError",
     # V2 enum exports
     "AuthenticationScheme",
     "FileType",
--- a/aixplain/v2/client.py
+++ b/aixplain/v2/client.py
@@ -1,13 +1,13 @@
 """Client module for making HTTP requests to the aiXplain API."""
 
-from typing import Any, Optional, Tuple, Union, List
+from typing import Any, Iterable, Optional, Tuple, Union, List, FrozenSet
 import logging
 import os
 import requests
 from requests.adapters import HTTPAdapter, Retry
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 
-from .exceptions import APIError
+from .exceptions import APIError, UntrustedURLError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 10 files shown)


elbadrashiny (approved):
**Verdict:** approve — the fix is correct, every load-bearing line is pinned by a test, and I reproduced all four security behaviours at runtime. Three non-blocking notes and one question for ops.

## Non-blocking

**1. The description understates the pre-fix behaviour (worth correcting on the ticket).**
The PR says `_is_poll_url` "matched host *and* path in one regex, so a poll-shaped URL on a foreign host silently fell through to being treated as **output media data**". That isn't what the old regex did. Its host group was optional and unanchored to any aiXplain domain:

```
^(?:https?://[^/?#]+)?/?(?:api/v1/data/[A-Za-z0-9_-]+|...)
```

`https://evil.example.com/api/v1/data/abc123` **matched**, so it was routed to `sync_poll` and the session-header credential went with it. The real pre-fix behaviour was worse than described — the key was sent, not withheld. The fix is right either way; the ticket record should say so.

**2. Three incidental widenings in `_is_poll_url`, all safe.** Comparing old vs new over a case matrix, the only behavioural deltas are:

| Input | Old | New |
|---|---|---|
| `…/api/v1/data/abc123/` (trailing slash) | no match | match |
| `HTTPS://PLATFORM-API.AIXPLAIN.COM/api/v1/data/abc` | no match | match |
| `//evil.example.com/api/v1/data/abc` | no match | match |

All three now route to `poll()` and hit the trust guard, which is the intended loud-failure path. The uppercase case is an incidental bug fix — that URL used to be misclassified as media data on a *legitimate* host.

**3. `upload_utils` gets a zero-trust session.** `RequestManager.create_session()` (`upload_utils.py:114-116`) calls `create_retry_session()` with no `trusted_origins`, so its allowlist is the empty `frozenset()` and *any* redirect strips the two credential headers. Harmless today — that path authenticates with `Authorization: token …`, which base `requests` already handles — and fail-closed is the right default. Worth knowing that the new allowlist covers `AixplainClient` only: the upload path's own credential and its `backend_url` POST are outside `ensure_trusted_url`.

## Question

`build_trusted_origins` trusts `base_url` + `model_url` + `pipeline_url` + `AIXPLAIN_TRUSTED_HOSTS` + the two hardcoded prod hosts. Is there any deployment where a legitimate poll URL comes back on a host outside that set — a regional endpoint, a CDN, an on-prem result host? If so those users hit `UntrustedURLError` on upgrade. `AIXPLAIN_TRUSTED_HOSTS` is the documented escape hatch and the error message names it, so this is a "confirm with ops" rather than a defect.

## Verified

- **Tests:** `pytest tests/unit` → **1604 passed, 4 skipped**. Merge-base baseline (`development` @ `42ed1e02`) → **1490 passed, 4 skipped**. No regressions; +114 new.
- **Mutation-tested the wiring, not just the guard** — each reverted individually, all go red:
  - `request_raw` using `resolve_url` instead of `ensure_trusted_url` → 3 failed
  - dropping `ensure_trusted_url(poll_url)` from `RunnableResourceMixin.poll` → 41 failed
  - `rebuild_auth` not stripping the credential headers → 4 failed
  - `UntrustedURLError` removed from the poll-loop re-raise tuple → 21 failed
  - credentials moved back onto the session → 4 failed
- **Runtime PoC** (local listeners, scratch dir, clone left pristine):
  - body-supplied foreign host → refused before any socket opens
  - `//evil.example.com/...` → refused after `urljoin` resolution
  - `https://platform-api.aixplain.com@evil.example.com/...` → refused
  - trusted → untrusted `302`: `x-api-key` present on hop 1, **absent** on hop 2, redirect still followed (`200`) — so the presigned-S3 flow keeps working
- **Callers enumerated:** both `create_retry_session` call sites; `_is_poll_url`'s single call site (`model.py:781`); `_auth_headers` cannot emit a `None` header because `__init__` raises unless exactly one key is set.
- **No v1 impact:** `create_retry_session` has no v1 callers, so `_AixplainSession` never touches the 76 v1 `_request_with_retry` sites.

**Not verified:** functional/integration suites (need live credentials); real behaviour against a deployed aiXplain backend; whether any production poll host falls outside the allowlist (the question above).

PR #1041, opened 2026-08-12
fix(v2): bound the remaining unbounded HTTP calls on the upload/file path (BUG-938)

aix-ahmet:
## Problem

`requests` has **no default timeout**. A call without an explicit `timeout=` waits forever, so a peer that completes the TCP handshake and then goes silent (LB draining, NAT blackhole, half-open socket) pins the calling thread with no upper bound — in `download_data`'s case, mid-`iter_content` with an open file handle.

`aixplain/v2/client.py` already grew a `default_timeout()` helper. This PR covers the callers on the upload/file path that it missed.

## Changes

| Location | Change |
|---|---|
| `v2/upload_utils.py` | `RequestManager.request_with_retry` `setdefault`s `default_timeout()` onto the forwarded kwargs. This is the single dispatch point for **both** upload legs (presigned-URL `POST` to the backend, S3 `PUT` of the file body), so one line bounds both. Callers keep their own bound if they pass `timeout=`. |
| `v2/code_utils.py` | `parse_code` and `parse_code_decorated` pass `default_timeout()` when fetching code from a URL. |
| `utils/file_utils.py` | `download_data` takes an optional `timeout` and defaults to `(10s connect, 300s read)`. |

### Configuration

Defaults are tunable through the **same** env vars as the v2 client, so one pair of knobs covers the whole SDK:

- `AIXPLAIN_HTTP_CONNECT_TIMEOUT` (default `10`)
- `AIXPLAIN_HTTP_READ_TIMEOUT` (default `300`)

A malformed or non-positive value is logged and ignored rather than silently removing the very bound it exists to configure.

### Why large transfers are safe

The read timeout bounds the **gap between chunks**, not the total transfer. A multi-GB download that keeps delivering bytes never trips it; only a stalled socket does.

## Tests

- **`tests/unit/v2/test_upload_timeouts.py`** — defaults applied, per-call override wins, env override honoured, unusable env values fall back; both upload legs verified to go through `request_with_retry` (so neither can bypass the default).
- **`tests/unit/utils/file_utils_test.py`** — the same matrix for `download_data`.
- **Blackhole tests** — a real loopback listener that completes the handshake and never answers, asserting bounded failure (`requests.exceptions.Timeout` / `FileUploadError`) instead of a hang. The shared `blackhole_url` fixture lives in the new `tests/unit/conftest.py` and skips where a sandbox forbids listening sockets.
- **`tests/unit/v2/test_no_unbounded_requests.py`** — an AST guard (modelled on the existing `test_no_v1_imports.py`) that fails if any HTTP dispatch in the guarded files regresses to an unbounded call. It credits `**kwargs` forwarding only when the enclosing function `setdefault`s a timeout, and includes anti-vacuity tests so the guard can't silently rot into always-passing.

### Verification

- `1272 passed` across `tests/unit`. The 32 collection errors are all a pre-existing missing `requests_mock` dependency in this environment, unrelated to this change.
- Blackhole tests confirmed passing against real sockets (11.5s wall clock, i.e. actually timing out rather than skipping).
- All pre-commit hooks pass, including `ruff`, `ruff format`, and `pytest-check`.

Written test-first: tests were red before the implementation landed.

### DIFF
--- a/aixplain/utils/file_utils.py
+++ b/aixplain/utils/file_utils.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import logging
 import os
 import re
 import requests
@@ -23,10 +24,48 @@ from aixplain.enums.license import License
 from aixplain.utils.request_utils import _request_with_retry
 from collections import defaultdict
 from pathlib import Path
-from typing import Any, Optional, Text, Union, Dict, List
+from typing import Any, Optional, Text, Tuple, Union, Dict, List
 from uuid import uuid4
 from urllib.parse import urljoin, urlparse
 from pandas import DataFrame
+
+logger = logging.getLogger(__name__)
+
+# Default (connect, read) timeout for the streaming download below. Without one,
+# ``requests`` blocks forever on a socket read: a blackholed download pins the
+# calling thread mid-``iter_content`` with an open file handle. The read timeout
+# bounds the gap between chunks, not the total download, so large files are fine.
+DEFAULT_DOWNLOAD_CONNECT_TIMEOUT = 10.0
+DEFAULT_DOWNLOAD_READ_TIMEOUT = 300.0
+
+
+def _download_timeout() -> Tuple[float, float]:
+    """Resolve the (connect, read) download timeout, honouring env overrides.
+
+    Uses the same ``AIXPLAIN_HTTP_CONNECT_TIMEOUT`` / ``AIXPLAIN_HTTP_READ_TIMEOUT``
+    names as the v2 client so one pair of variables tunes the whole SDK. A
+    malformed or non-positive value must not silently remove the bound it exists
+    to configure, so it is logged and the default is kept.
+    """
+
+    def _resolve(name: str, default: float) -> float:
+        raw = os.getenv(name, "").strip()
+        if not raw:
+            return default
+        try:
+            value = float(raw)
+        except ValueError:
+            logger.warning(f"Ignoring non-numeric {name}={raw!r}; using {default}s")
+            return default
+        if value <= 0:
+            logger.warning(f"Ignoring non-positive {name}={raw!r}; using {default}s")
+            return default
+        return value
+
+    return (
+        _resolve("AIXPLAIN_HTTP_CONNECT_TIMEOUT", DEFAULT_DOWNLOAD_CONNECT_TIMEOUT),
+        _resolve("AIXPLAIN_HTTP_READ_TIMEOUT", DEFAULT_DOWNLOAD_READ_TIMEOUT),
+    )
 
 
 def save_file(download_url: Text, download_file_path: Optional[Union[str, Path]] = None) -> Union[str, Path]:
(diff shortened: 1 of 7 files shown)


elbadrashiny (approved):
**Verdict:** approve — small, well-scoped, and the tests are the good kind (real sockets, not mocks). One non-blocking finding that's worth more than the PR itself: the sibling call in the same file is still unbounded, and the new guard structurally cannot see it.

## Non-blocking (but high leverage)

**`save_file()` — same file, same file-download path, inside a file this PR's guard covers — is still unbounded.**

```python
# aixplain/utils/file_utils.py, ~15 lines above the fixed download_data
r = _request_with_retry("get", download_url)
```

`_request_with_retry` (`aixplain/utils/request_utils.py`) builds a bare session and passes `**params` straight through with no timeout default, so `save_file` has exactly the failure mode the PR describes: a peer that completes the handshake and goes silent pins the thread forever.

And the AST guard can't catch it. `test_no_unbounded_requests.py` lists `aixplain/utils/file_utils.py` in `GUARDED_FILES`, but only flags `ast.Attribute` calls whose receiver matches `DISPATCH_RECEIVERS = ("requests", "session")`. `_request_with_retry("get", url)` is a plain `ast.Name` call, so it's invisible — the guard reports that file clean while an unbounded dispatch sits in it. That's the specific shape this kind of guard is worst at: it gives false assurance about the exact file it covers.

The high-value version of this fix is one line in the helper, because it bounds everything at once:

```python
def _request_with_retry(method, url, **params):
    params.setdefault("timeout", ...)   # bounds all 76 call sites
```

`_request_with_retry` has **76 call sites across 29 files, and not one passes a `timeout=`** — so the entire v1 HTTP surface is currently unbounded. That's out of this PR's stated scope ("the callers on the upload/file path that it missed"), and I'm not asking for it here. But it's worth a ticket, and it would be a shame to leave `save_file` behind when it's in the diff's own file. At minimum, extend `_is_unbounded_dispatch` to treat calls to known local HTTP helpers as dispatches, so the guard stops reporting a false clean.

## Verified

- **Tests:** `pytest tests/unit` → **1526 passed** (merge-base `development` @ `42ed1e02`: **1490 passed**). No regressions; +36 new.
- **The blackhole tests genuinely time out rather than skipping** — 19 passed in 11.55s wall clock against real loopback listeners. That's the difference between testing a timeout and asserting on a mock, and it's the right call here.
- **Read the timeout resolution:** `_download_timeout()` reuses the v2 client's `AIXPLAIN_HTTP_CONNECT_TIMEOUT` / `AIXPLAIN_HTTP_READ_TIMEOUT` names, and a malformed or non-positive value logs and keeps the default rather than silently dropping the bound — the correct direction for a safety default.
- **`request_with_retry`'s `setdefault` is the right insertion point:** it's the single dispatch for both upload legs (presigned `POST` at `upload_utils.py:162`, S3 `PUT` at `:181`), so one line bounds both and a caller passing `timeout=` still wins. Confirmed both legs route through it.
- **Signature change is backward compatible:** `download_data(url_link, local_filename=None, timeout=None)` — new param is last and defaulted; no non-test caller in the tree passes positionally past `local_filename`.
- **No circular import risk:** `code_utils` imports `default_timeout` from `.client` at module scope (client imports only `.exceptions`), while `upload_utils` keeps the lazy in-function import, matching its existing `create_session` pattern.
- The read timeout bounding inter-chunk gaps rather than total transfer is correct as described, so large downloads aren't capped.

**Not verified:** real-world behaviour against S3 for a multi-GB upload; the functional suites.

**One cross-PR note:** once #1045 lands, the presigned-URL `POST` at `upload_utils.py:162` will both gain this default read timeout *and* lose its 500/502/503/504 retry (that PR drops `POST` from `allowed_methods`). Those two are individually fine but land on the same call — flagged on #1045 as well.

PR #1042, opened 2026-08-12
fix(v2/rlm): stop baking TEAM_API_KEY into RLM sandbox source (BUG-936, SEV1)

aix-ahmet:
## Problem (SEV1)

`RLM`'s recursive mode injected an `llm_query` helper into the code-execution sandbox with the caller's **account-level `TEAM_API_KEY` interpolated straight into the generated source**:

```python
_headers = {"x-api-key": "<TEAM_API_KEY>", "Content-Type": "application/json"}
```

The sandbox is a third-party execution service with outbound network access, and the code running inside it is written by an orchestrator model that is steered by the analysed context. Any prompt injection in that context could recover the key — `llm_query.__code__.co_consts`, `inspect.getsource`, etc. — and POST it anywhere. The account key grants full platform access, so this is credential compromise, not just data leakage.

Two related gaps: the sandbox session was never torn down (uploaded context and state outlived the run), and SDK-generated setup code shared one session with model-emitted code.

## Approach

Rather than *hide* the key inside the sandbox, **remove it**. The worker call moves SDK-side, where the credential already lives.

### 1. Credential-free `llm_query`

The prelude is now a module-level constant with **no interpolation at all** — no caller value, least of all a key, can reach sandbox-executed source. It submits prompts on stdout behind a marker; the SDK strips them out of the output, answers them in-process (concurrently, deduplicated by prompt hash), and pushes answers back as a plain JSON string literal.

Because answers are fetched *between* REPL turns, a batch resolves over two turns:

| Turn | Behaviour |
|---|---|
| A | `llm_query(p)` submits `p`, returns a `PENDING` placeholder |
| B | Re-run the **same block** — every prompt returns its real answer, batch answered in parallel |

The system prompt and REPL example teach this shape explicitly, and placeholders are barred from every exit: `FINAL(...)` is rejected with a retry prompt, `FINAL_VAR(...)` returns `None`, and the forced final answer strips them.

- `max_iterations` default **10 → 12** to absorb the extra submit turn.
- Credit tracking moves into `_worker_call` — worker spend is now counted SDK-side instead of read back from a sandbox global (which also removes a trust dependency on a sandbox-resident counter).

### 2. Session isolation

Every run now gets two sessions with fresh UUIDs:

- **`setup`** — SDK-generated code only. Orchestrator-emitted code can never reach it, making it the enforced home for any future bootstrap step that must handle a credential.
- **`repl`** — the context, the prelude, and every model-emitted block.

`_run_sandbox` takes a **required keyword-only `session`**, so every call site has to state which side of the trust boundary it is on; emitted code is hard-wired to `repl`. A cheap preflight probe keeps the split exercised rather than merely declared, and turns a dead sandbox into an early, clearly-attributed failure.

### 3. Teardown

Both sessions are closed in a `finally` that covers the success, exception, and timeout paths. The close action name is discovered at runtime (`Tool.run` validates `action` against advertised actions, so it can't be hard-coded) and falls back to wiping session globals when the tool exposes none — a rejected close action still wipes rather than leaking the uploaded context. `close()` is idempotent and never raises, so a teardown failure can't turn a good run bad or mask an in-flight exception. `RLM` is also usable as a context manager.

### 4. Defence in depth

- **Tripwire** — SDK-generated code containing a credential raises `ResourceError` (a BUG-936 regression guard); emitted code containing one is refused without killing the run.
- **Redaction** — sandbox stdout, `FINAL_VAR` values, final answers, and error messages are scrubbed of credential values. Covers the resolved client key *and* `TEAM_API_KEY` / `AIXPLAIN_API_KEY` from the environment, so the tripwire still fires if a key arrives by a path this class doesn't own. Redaction happens before truncation so a secret can't survive as a fragment.
- **Download URL** — the temp context link is a bearer capability; it's now unbound after load on both the upload and URL fast paths.
- **`FINAL_VAR` names** are model-controlled and interpolated into generated code, so non-identifiers are rejected instead of executed.
- **Answer injection** carries answers as `repr(json.dumps(...))` — attacker-influenced answer text can't break out into executable code.

## Tests

53 new tests (90 total in `rlm_test.py`, all passing):

- **No credential in the sandbox** — asserts the key appears in *no* sandbox payload; asserts module source never interpolates a key into generated code.
- **Recovery probes** — `co_consts`, `getsource`, and `os.environ` dumps can't recover the key; echoed secrets and failure messages are redacted.
- **Session split** — sessions are distinct, emitted code only ever hits `repl`, setup state is unreachable from `repl`, unstarted/unknown sessions are rejected.
- **Teardown** — on success, on exception, on setup failure; failed close still wipes; wipe fallback without a close action; idempotency; context manager.
- **Protocol** — marker stripping, malformed and truncated payloads, same-line prefixes, two-turn resolution, hash dedup, keying on the full prompt while truncating worker input, quote/backslash round-trips, marker never reaching the transcript, stranded-prompt resubmission, notes surviving a very long block output.
- **Bootstrap & final answer** — URL unbinding on both paths; `FINAL_VAR` identifier and placeholder rejection; `FINAL` placeholder rejection with run continuation; forced-final stripping.

```
90 passed in 0.64s
```

Full pre-commit suite (ruff, ruff format, pytest-check) passes. No other module touched.

## Compatibility

`llm_query(prompt)` keeps its signature and its 8192-token output budget, so answer length is unchanged. The behavioural change is the two-turn resolution — the orchestrator is instructed on it directly, and `max_iterations` was raised to match. `llm_query_request([...])` is available to submit a whole batch in one call.

### DIFF
--- a/aixplain/v2/rlm.py
+++ b/aixplain/v2/rlm.py
@@ -2,12 +2,18 @@
 
 Orchestrates long-context analysis via an iterative REPL sandbox. The
 orchestrator model plans and writes Python code to chunk and explore a large
-context; a worker model handles per-chunk analysis via ``llm_query()`` calls
-injected into the sandbox session.
+context; a worker model handles per-chunk analysis via ``llm_query()``.
+
+``llm_query`` is credential-free inside the sandbox: the injected helper submits
+prompts on stdout and the SDK answers them in-process, so no API key is ever
+materialised in sandbox-executed source (BUG-936). SDK-generated setup code and
+model-emitted code run in separate sandbox sessions, and both are torn down when
+the run ends.
 """
 
 __author__ = "aiXplain"
 
+import hashlib
 import json
 import logging
 import os
@@ -20,7 +26,7 @@ import uuid
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass, field
 from dataclasses_json import dataclass_json, config as dj_config
-from typing import Any, Dict, List, Optional, Union, TYPE_CHECKING
+from typing import Any, Dict, List, Optional, Tuple, Union, TYPE_CHECKING
 
 from .resource import BaseResource, Result
 from .mixins import ToolableMixin, ToolDict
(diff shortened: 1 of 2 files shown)


elbadrashiny (approved):
**Verdict:** approve — the SEV-1 is real, the redesign removes the credential from the sandbox entirely rather than trying to sanitise it, and the invariant is defended three independent ways, all of which I verified fire. Two non-blocking notes.

The old shape was as bad as the ticket says:

```python
_headers = {{"x-api-key": "{self.context.api_key}", "Content-Type": "application/json"}}
```

— the account-level `TEAM_API_KEY` as a string literal in source executed inside the sandbox, so any prompt injection in the analysed context could recover it (`llm_query.__code__.co_consts`) and exfiltrate it. Replacing the in-sandbox HTTP call with a stdout request/response protocol is the right fix: it removes the secret rather than hiding it, which is the only version of this that survives the next prompt-injection technique.

## Non-blocking

**1. The two-turn `llm_query` protocol is a real behavioural change for the orchestrator, and its risk is in the prompt, not the code.** A first call now returns `_LLM_PENDING_SENTINEL` and the answer arrives on the next turn, so a batch costs two iterations (acknowledged in the docstring). The code side is bounded and defensive — 256 prompts/batch, 512 markers/block, 400k chars/prompt, malformed payloads skipped with a warning, and the sentinel is barred from reaching `RLMResult.data`. What I can't check by reading is whether the model reliably follows "re-run the SAME block next turn" instead of treating the placeholder as an answer. That's an eval question and worth a few real runs on recursive-mode queries before this is relied on.

**2. The sandbox's own context download is unbounded:** `__requests.get(_url, stream=True)` in the URL fast path (`rlm.py:1013`) passes no `timeout`, so a stalled presigned URL hangs the sandbox block. Blast radius is bounded by the sandbox's own limits rather than a caller thread, so it's much milder than the cases #1041 fixes — but it's the same shape and one `timeout=` away.

**3. One-off test flake, reported for completeness, not attributed to this PR.** My first full-suite run on this branch showed `tests/unit/agent/agent_factory_utils_test.py::test_build_agent_success_cases[agent_with_tools]` failing. It did not reproduce in 13 subsequent runs (in isolation, whole-file, and full-suite across `PYTHONHASHSEED=0..5` on both this branch and `development`). Most likely stale `__pycache__` from my switching branches in one worktree. Flagging the test name only so it can be watched.

## Verified

- **No credential reaches the sandbox — traced, not assumed.** Every `api_key` reference in `rlm.py` classified: `:686` is a docstring example, `:925` is `_secret_values()` reading its own key to *detect* it, `:1068` is `FileUploader(api_key=…)` running **in the SDK process**. The sandbox receives only a temp download URL. The prelude imports nothing but `json` and `hashlib`. The two `context_code` f-strings interpolate only `repr(context)`.
- **The guard is on the shared chokepoint, which is the part that matters.** `_assert_no_secrets(code)` sits inside `_run_sandbox` (`rlm.py:1155`), so it covers SDK-generated *and* model-emitted code — not just the orchestrator path.
- **Mutation-tested all three controls; each goes red:**
  - drop `_assert_no_secrets` from `_run_sandbox` → 1 failed
  - make `_contains_secret` always return `False` → 3 failed
  - stop redacting secrets from sandbox output → 2 failed
- **Reintroduced the original bug and confirmed it is caught.** Interpolating the *real* credential into `context_code` (exactly BUG-936's shape) fails 4 tests — including a dedicated static invariant test, `TestNoCredentialInSandbox::test_module_source_never_interpolates_a_key_into_generated_code` — and the runtime tripwire raises:

  ```
  ResourceError: RLM: refusing to send code containing an API key to the sandbox. This is the BUG…
  ```

  and `_redact()` turns the key into `***REDACTED***`. (My first attempt at this mutation injected a *fake* literal and everything stayed green — correctly, since it isn't the secret. Worth knowing the guard is value-based, not pattern-based: it catches the real key, not a key-shaped string.)
- **`_MIN_SECRET_LEN = 8`** is a necessary detail — without it a short/degenerate env value would match every string and redact everything.
- **Tests:** `tests/unit/rlm_test.py` → **90 passed**; full `tests/unit` → **1538 passed** (baseline `development`: 1494). No regressions; +44 new.
- **Session separation** is enforced structurally: `session` is a required keyword on `_run_sandbox` and `_session_id_for` raises on an unknown or unstarted session, so model-emitted code cannot reach the setup session.

**Not verified:** any of this against a real sandbox tool — every sandbox interaction here is mocked, so the protocol is proven at the unit level only; the credit accounting move from the in-sandbox `_total_llm_query_credits` to the SDK's `self._used_credits` (I did not check billing totals match the old path); orchestrator behaviour under the new two-turn protocol (note 1).

PR #1043, opened 2026-08-12
ENG-3431: unit test integrity — coverage gate, credential-free collect, v2 effect assertions

aix-ahmet:
## Summary

The unit suite could report green while checking very little. This PR closes four independent holes and adds the gates that keep them closed.

| Hole | Before | After |
|---|---|---|
| Assertion-free tests | Counted as passes | Collection fails, with an allowlist that requires a written reason |
| Silent `importorskip` | Whole plotting modules skipped on an undeclared package | `plotly` declared; a guard test fails if a new silent skip appears |
| Coverage | `--source=.` (inflated by the `tests/` package), never gated | Scoped to `aixplain`, gated in CI at 62% |
| Credentials | Unit suite couldn't be **collected** without `TEAM_API_KEY` | Import never raises on a missing key; a subprocess test proves it |

### Coverage gate
- Coverage scope moved to `[tool.coverage.run]` in `pyproject.toml` (`source = ["aixplain"]`). `--source=.` was measuring the test files themselves — `tests/` ships as a package via `[tool.setuptools.packages.find]` — and inflating the headline number. The report now has **zero `tests/` rows**.
- New `unit-coverage` CI job runs the unit suite and enforces `coverage report --fail-under=$COVERAGE_FLOOR`.
  - Floor is **62**, two points below the 65% measured on this branch, to absorb ordering/version jitter. Ratchet it up in the workflow `env` block as coverage improves.
  - The threshold is passed on the command line rather than set as `fail_under` in `pyproject.toml`, so the gate is visible in the workflow and a local or pre-commit run is never blocked mid-refactor.
- `unit` removed from the `setup-and-test` matrix so it isn't run twice.
- CI and the pre-commit hook install `coverage[toml]`, not plain `coverage`: on Python < 3.11 coverage can't read `pyproject.toml` without the extra and aborts with *"Can't read 'pyproject.toml' without TOML support"* before a single test runs.

### Assertion guard (`tests/conftest.py`)
`pytest_collection_modifyitems` now raises `pytest.UsageError` if any collected `tests/unit` test has no assertion. Design notes:

- Inspects the **resolved function's** source, not a file-level AST sweep. A sweep false-positives on nested `test_`-named helpers inside real tests (`utility_tool_decorator_test.py` has exactly that).
- Follows same-module and `self.` helpers up to 3 levels, because delegating to a shared `_assert_payload(...)` is an ordinary, correct way to write a test. Since the guard takes the *whole suite* down when it fires, erring toward a false negative is the right trade.
- Recognises non-`assert` assertions: `pytest.raises`/`warns`, the `mock.assert_*` family, `unittest` `assertEqual` &c., plus any project-local name starting with `assert`.
- Unreadable or dynamically generated functions get the benefit of the doubt rather than failing the run on a parsing limitation.
- Runs **last**, so it only sees items that survived the existing version filters.
- Scoped to `tests/unit` on purpose — functional tests hit a live backend and some legitimately assert only by not raising. Widening is a follow-up, not a silent side effect.

### Credential-free collection (`aixplain/utils/config.py`)
`validate_api_keys()` ran at import time and raised when no key was set, so the unit suite couldn't even be collected without a credential. Split into:

- `_normalize_api_keys()` — runs at import, **never raises on absence**. Normalization has to stay eager because ~40 v1 call sites bind `config.TEAM_API_KEY` as a default argument value, evaluated once at import, so the `AIXPLAIN_API_KEY -> TEAM_API_KEY` copy must happen before those modules load.
- `validate_api_keys()` — kept, now normalize-then-require, for callers that want the eager all-or-nothing check.

Conflicting keys still fail loudly and early: that can only be intentional misconfiguration, and it can't fire when no key is set, so it never blocks collection.

> [!NOTE]
> **Accepted regression, called out in the docstring.** v1 has no runtime credential check — the `@check_api_key` decorator in `aixplain/v1/decorators` is applied to no call site, so the import-time raise removed here was the only credential error v1 ever produced. A v1 caller with no key now gets a backend 401 instead of an actionable message. v1 is deprecated, so this is accepted rather than fixed here.

New `tests/unit/utils/test_credential_free_collection.py` collects the entire unit suite in a subprocess with no credential in the environment and `load_dotenv` neutered via `sitecustomize` (imported by `site` at startup, before anything can read the environment) — plus a test that the neutering itself works, so the guard can't pass vacuously.

### Silent skips
- `plotly>=5.18.0` declared in the `test` extra and in the pre-commit hook deps. The `agent_evaluator` / `eval_experiment` plotting tests were `pytest.importorskip("plotly")`-ing themselves away and had **never run**.
- New `tests/unit/test_dependency_declarations.py` asserts every `importorskip` target in the suite is a declared dependency *and* is actually importable.

### Tests that asserted nothing, fixed
- **`image_upload_test.py`** — real bug in the test: `ModelFactory.list_functions(config.TEAM_API_KEY)` passed the key as the **`verbose` positional**, not `api_key`. It only passed because a non-empty key is truthy, so the test silently depended on a credential being set. Now `list_functions(True, api_key=...)`.
- Gave the "should not raise" tests in `utility_test.py`, `test_api_key.py`, `test_agent_budget.py` and `test_v2_agent_v3_compat.py` real assertions on the *effect* — request path and count, purity of validation (valid input survives untouched, nothing fabricated), and absence of a `DeprecationWarning` verified via `record=True` rather than `simplefilter("error")`.

### New v2 effect-based coverage
- **`test_resource_paths.py`** — asserts the backend path each v2 resource actually talks to for `get`/`search`/`create`/`delete`, plus Skill download / file upload / folder upload. The path table is kept as literals on purpose: it's the specification, not a mirror of the constant under test.
- **`test_upload_headers.py`** — asserts the presigned-URL request carries the Authorization token, the S3 PUT carries `Content-Type` and **leaks no credential**, and temp vs permanent uploads hit their respective endpoints.

## Verification

Ran locally in a clean venv with the `test` extra, no `TEAM_API_KEY` / `AIXPLAIN_API_KEY` / `BACKEND_URL` set:

```
$ coverage run -m pytest tests/unit -q
1528 passed in 16.05s

$ coverage report --fail-under=62   -> exit 0   (TOTAL 21164 stmts, 65%)
$ coverage report --fail-under=99   -> exit 2   (gate fails as expected)
$ coverage report | grep -c '^tests/'  -> 0     (test files no longer measured)
```

Assertion guard confirmed in the failing direction with a temporary probe:

```
ERROR: Tests with no assertion (add one, or allowlist it with a written reason
in ASSERT_FREE_ALLOWLIST in tests/conftest.py):
  tests/unit/zz_guard_probe_test.py::test_probe_has_no_assertion
```

All pre-commit hooks pass, including `pytest-check` under the updated hook dependencies.

## Reviewer notes

- The `COVERAGE_FLOOR` value (62) is the one knob worth arguing about — it's in the workflow `env` block, deliberately visible.
- The accepted v1 no-credential-message regression in the note above is the only behaviour change outside tests and CI.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,15 +8,65 @@ on:
       - test
   workflow_dispatch:
 
+env:
+  # Coverage floor for the unit suite, enforced by the `unit-coverage` job.
+  # Measured at 64% on `development` @ 42ed1e02; set two points below to absorb
+  # ordering/version jitter. Ratchet it up here as coverage improves.
+  COVERAGE_FLOOR: "62"
+
 jobs:
+  # The unit suite is deliberately credential-free: no TEAM_API_KEY, no
+  # BACKEND_URL. If this job starts failing on a missing key, the regression is
+  # in the code under test (something validating credentials at import time),
+  # not in this workflow.
+  unit-coverage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
+          cache: 'pip'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          # `coverage[toml]` (not plain `coverage`): the coverage settings live in
+          # pyproject.toml, and on Python < 3.11 coverage needs the `tomli` extra to
+          # read it. Without the extra it aborts with "Can't read 'pyproject.toml'
+          # without TOML support" before a single test runs.
+          pip install ".[test]" "coverage[toml]"
+
+      - name: Run unit tests with coverage
+        run: coverage run -m pytest tests/unit
+
+      - name: Enforce coverage floor
+        run: coverage report --fail-under=${{ env.COVERAGE_FLOOR }}
+
+      - name: Write coverage.xml
+        if: always()
+        run: coverage xml
+
+      - name: Upload coverage report
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: coverage-xml
+          path: coverage.xml
+
   setup-and-test:
     runs-on: ubuntu-latest
     timeout-minutes: 45  # Global timeout fallback
     strategy:
       fail-fast: false
       matrix:
+        # 'unit' is absent on purpose: it runs in the credential-free
+        # `unit-coverage` job above, so it is not executed twice.
         test-suite: [
-          'unit',
           'file_asset',
           'data_asset',
           'model',
@@ -28,9 +78,6 @@ jobs:
           'v2',
         ]
         include:
-          - test-suite: 'unit'
-            path: 'tests/unit'
-            timeout: 45  # Tweaked timeout for each unit tests
           - test-suite: 'file_asset'
             path: 'tests/functional/file_asset'
             timeout: 45
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,18 +22,24 @@ repos:
     hooks:
       - id: pytest-check
         name: pytest-check
-        entry: coverage run --source=. -m pytest tests/unit
+        # Coverage scope lives in pyproject.toml ([tool.coverage.run]); `--source=.`
+        # measured the test files too and inflated the figure. The floor is only
+        # enforced in CI, so a local run is never blocked mid-refactor.
+        entry: coverage run -m pytest tests/unit
         language: python
         pass_filenames: false
         types: [python]
         always_run: true
         additional_dependencies:
-          - coverage
+          # `[toml]` extra required to read [tool.coverage.run] from pyproject.toml
+          # on Python < 3.11; plain `coverage` hard-errors there.
+          - coverage[toml]
           - pytest
           - pytest-mock
           - pytest-rerunfailures
           - pytest-xdist
           - requests-mock
+          - plotly>=5.18.0  # keeps the plotting tests from silently skipping here too
           - Babel>=2.12.0
           - requests>=2.1.0
           - tqdm>=4.1.0
--- a/aixplain/utils/config.py
+++ b/aixplain/utils/config.py
@@ -28,23 +28,32 @@ AIXPLAIN_API_KEY = os.getenv("AIXPLAIN_API_KEY", "")
 ENV = "dev" if "dev" in BACKEND_URL else "test" if "test" in BACKEND_URL else "prod"
 
 
-def validate_api_keys():
-    """Centralized API key validation function - single source of truth.
+def _normalize_api_keys():
+    """Reconcile the two API key environment variables. Never raises on absence.
 
-    This function handles all API key validation logic:
-    1. Ensures at least one API key is provided
-    2. Prevents conflicting API keys
-    3. Auto-normalizes AIXPLAIN_API_KEY to TEAM_API_KEY if needed
+    Importing this module must stay side-effect-safe when no credential is set:
+    the unit test suite has to be collectible without one. Only *normalisation*
+    happens eagerly, because roughly 40 v1 call sites bind ``config.TEAM_API_KEY``
+    as a default argument value, which is evaluated once at import time - so the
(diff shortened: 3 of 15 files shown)


elbadrashiny (approved):
**Verdict:** approve — I reproduced the headline defect and the fix, and both negative claims in the description hold up under a control query. One non-blocking suggestion about the accepted v1 regression.

The premise is not overstated: on `development` @ `42ed1e02`, `pytest tests/unit` with no credential set **cannot be collected at all** — 27 collection errors, `Interrupted: 27 errors during collection`. On this branch the same command gives **1523 passed**. That is a real hole, and it was hiding the other three.

## Non-blocking

**The accepted v1 regression is the one user-visible cost, and it has a cheap mitigation.** The docstring is right that `@check_api_key` is applied nowhere, so removing the import-time raise leaves v1 with no credential check at all — a new user with no key now gets a backend `401` instead of the message pointing at the setup docs. Since v1 is still the path most existing code takes (and #1044 only *announces* the deprecation, it doesn't remove v1), consider one of:

```python
# aixplain/v1/__init__.py
from aixplain.utils.config import check_api_keys_available
check_api_keys_available()
```

or actually applying the existing `@check_api_key` decorator to the v1 factory entry points. Either keeps collection credential-free (v1 isn't imported by `import aixplain` — verified, 0 v1 modules load) while restoring the actionable error. Not blocking, and clearly a conscious trade as written.

Smaller notes:
- The workflow comment says the floor is "two points below the 64% measured on `development`", the PR body says "two points below the 65% measured on this branch". I measure **65%** on this branch. Both land on 62, so this is only a comment tidy-up.
- The assertion guard raises `UsageError`, so one assertion-free test takes the *whole* suite down for everyone, and the escape hatch is an allowlist edit. That's the intended design and the false-positive handling is careful (below) — noting it so nobody is surprised the first time it fires.

## Verified

- **Headline claim, both directions:**
  - `development`, no credential → **27 collection errors**, suite unrunnable.
  - this branch, no credential → **1523 passed**; with `plotly` installed as `.[test]` provides → **1528 passed, 0 skipped**.
- **Silent skips confirmed closed:** skip count goes 4 → **0** once `plotly` is present, i.e. the `agent_evaluator` / `eval_experiment` plotting tests now actually execute. (My first run showed 1 failure — `test_importorskip_targets_are_actually_importable[plotly]` — because my venv predated the new `test` extra. Installing `plotly` cleared it; that's the guard doing its job, not a PR defect.)
- **Coverage:** `coverage run -m pytest tests/unit` → `coverage report` = **65% TOTAL**, and **zero `tests/` rows** in the report. Both claims hold.
- **Both negative claims control-verified** (the search was first shown to find something it should):
  - `@check_api_key` — the definition exists (`aixplain/v1/decorators/api_key_checker.py:6`), and the only `@check_api_key` occurrence in the tree is inside that file's own docstring. Applied to no call site: **confirmed**.
  - `ModelFactory.list_functions(cls, verbose=False, api_key=None)` (`model_factory/__init__.py:258`) — so `list_functions(config.TEAM_API_KEY)` really did pass the key as the `verbose` positional. Real test bug, correctly diagnosed.
- **All four gates mutation-tested, and the false-positive guard too:**
  - added an assertion-free unit test → collection aborts, whole suite down ✅
  - same test with the assertion delegated to a `_assert_payload()` helper → **1529 passed**, no false positive ✅
  - added an `importorskip("nonexistent_undeclared_pkg")` → `test_dependency_declarations` fails on both the declared-dependency and importable checks ✅
  - restored import-time `validate_api_keys()` → `test_credential_free_collection` fails both tests ✅
- **Read the guard implementation**, not just its tests: resolves the function and parses its own source (so nested `test_`-named helpers don't false-positive), follows same-module and `self.` helpers 3 deep, recognises `pytest.raises`/`warns`, the `mock.assert_*` family, `unittest` assertions and any `assert*`-prefixed project helper, and gives unparseable functions the benefit of the doubt. It runs last, after the version filters.
- `unit` removed from the `setup-and-test` matrix is not a coverage loss — `main.yaml` triggers only on push to `main`/`test`, so the unit suite's real per-PR gate is the `pre-commit` hook either way, and that still runs it.

**Not verified:** the `unit-coverage` CI job itself (can't execute GitHub Actions here) — in particular that `coverage[toml]` resolves on the 3.9 runner; the functional suites.

PR #1044, opened 2026-08-12
ENG-3430: execute v1 deprecation — import-time warnings, MIGRATION.md, reconcile sunset date

aix-ahmet:
## Summary

The docs said SDK v1 was supported "until **August 1, 2026**" — a date that had already passed, while the SDK itself said nothing at all at runtime. This PR executes the deprecation properly: an import-time notice, a real migration guide, and one honest sunset date across every doc.

**New sunset date: February 1, 2027** (`2027-02-01`). It is deliberately not "soon" — eight v1 factories still have no v2 equivalent, and the date is contingent on closing those gaps first.

### 1. Import-time deprecation warning (`aixplain/_compat.py`)

`_compat.py` is where the legacy import redirector already lives, and **every** v1 route funnels through it (`aixplain.modules`, `aixplain.factories`, `aixplain.enums`, `aixplain.decorators`, `aixplain.base`, `aixplain.processes`). Hooking `create_module`/`load_module` there covers the whole v1 surface **without touching a single v1 module**. `aixplain/v1/__init__.py` warns directly, because submodule imports (`import aixplain.v1.factories`) bypass a parent package's `__getattr__`.

```text
'aixplain.factories' is part of the deprecated aiXplain SDK v1 and will be removed on
February 1, 2027 (2027-02-01). Migrate to the v2 API ('from aixplain import Aixplain').
Migration guide: https://github.com/aixplain/aiXplain/blob/main/MIGRATION.md.
Set AIXPLAIN_SUPPRESS_V1_DEPRECATION=1 to silence this notice.
```

Four design decisions worth reviewing:

- **`AixplainV1DeprecationWarning(DeprecationWarning)`** — subclassing means `-W error::DeprecationWarning` and `pytest.warns(DeprecationWarning)` both match, while users still get a dedicated category they can silence on its own.
- **Warn once per process for all of v1** — not once per module, not once per prefix. A single `import aixplain.modules` fans out to ~160 redirected imports, and v1 imports *itself* through the legacy paths (`aixplain/v1/enums/function.py` does `from aixplain.modules.model import Model`). Anything finer would make the SDK warn about itself dozens of times, which is precisely the noise that drives people to blanket-suppress `DeprecationWarning`. Remediation is a codebase-wide grep, not a per-import fix.
- **Computed `stacklevel`** — warnings raised inside a `MetaPathFinder` sit under a *variable* number of `importlib._bootstrap` frames (measured 4–7, depending on import depth), so any fixed `stacklevel` would blame the import machinery. We walk to the first frame outside importlib/aixplain, skipping bootstrap frames the way CPython's own `is_internal_frame` does, so the notice points at the user's actual `import` line.
- **The notice is made visible** — Python shows `DeprecationWarning` only when attributed to `__main__`, which is exactly how the first sunset date managed to pass in silence. We insert a `default` filter at index 0 (appending doesn't work: `warnings.filters` is first-match-wins and the stdlib's blanket `ignore` entry already matches). **This is guarded** — if `sys.warnoptions` is non-empty (`-W` or `PYTHONWARNINGS`), we leave the filters completely alone.

  <details><summary>The one deliberate sharp edge</summary>

  A `warnings.simplefilter("ignore")` issued *before* `import aixplain` leaves `sys.warnoptions` empty and so gets overridden by our filter. Judged the right trade for a sunset notice that already expired unseen once — there are three documented off switches (env var, `-W`/`PYTHONWARNINGS`, or any filter set after the import, since `filterwarnings` also inserts at index 0). The behaviour is pinned by tests rather than left implicit.
  </details>

Off switch: `AIXPLAIN_SUPPRESS_V1_DEPRECATION=1`.

### 2. `MIGRATION.md` (new)

Factory-by-factory v1 → v2 map with before/after code for each, how to find your exposure (`grep` one-liner + `python -W error::DeprecationWarning -c 'import your_app'`), and how to silence the notice.

Critically, it includes an explicit **"No v2 equivalent yet"** section naming the eight factories that have nowhere to go — `IndexFactory`, `PipelineFactory`, `BenchmarkFactory`, `CorpusFactory`, `DataFactory`, `DatasetFactory`, `FinetuneFactory`, `WalletFactory` — plus the model-onboarding helpers on `ModelFactory`. These gate the removal date. We're not removing capability that has nowhere to land.

### 3. Sunset date reconciled

`aixplain._compat.V1_REMOVAL_DATE` is the **single source of truth**. `tests/unit/test_v1_deprecation_docs.py` fails if any hand-written doc disagrees with it — so the constant changes first and the docs follow, never the reverse. That's what stops the date from silently rotting again.

Updated: `README.md`, `README.ar.md`, `docs/README.md`, `docs/README.ar.md`, `AGENTS.md`, `AGENTS.ar.md`.

The `AGENTS.md` note also spells out that deprecation **does not** license removing v1 code, pointing back to the existing V1 and Compatibility Rules.

## Verification

- **`1596 passed, 4 skipped`** — full unit suite, clean env and with `AIXPLAIN_SUPPRESS_V1_DEPRECATION=1` exported. 106 of those are new.
- Acceptance criteria, exit codes captured:

  | Command (under `-W error::DeprecationWarning`) | Exit |
  | --- | --- |
  | `import aixplain.v1` | `1` ✅ |
  | `import aixplain.factories` | `1` ✅ |
  | `from aixplain import Aixplain` | `0` ✅ |
  | `import aixplain.v1` with suppression env var | `0` ✅ |

  The v2-only case matters most: `import aixplain` loads **zero** v1 modules, so v2 users never see the notice.

## Reviewer notes

- No v1 module was modified — only `_compat.py` (the redirector) and `aixplain/v1/__init__.py` (docstring + one call).
- No behaviour change beyond the warning: every legacy import still resolves to the same objects.
- The stacklevel walk and the filter-insertion guard are the two subtlest pieces; both are commented in-place with the reasoning and pinned by tests.


elbadrashiny (approved):
**Verdict:** approve — I re-ran the whole acceptance table independently and every row reproduces, including the two things most likely to be subtly wrong: that a v2-only import stays silent, and that the computed `stacklevel` blames the user's own `import` line. One design decision worth an explicit team yes.

## Non-blocking

**1. `_make_notice_visible()` has a library mutate process-global `warnings.filters` at import time.** That's normally a thing libraries shouldn't do, so it deserves a conscious yes rather than passing unremarked. In its favour: the filter is scoped to `AixplainV1DeprecationWarning` (not a blanket `default`), it's skipped entirely when `sys.warnoptions` is non-empty, and there are three documented off-switches. I verified all three behaviours, including the sharp edge the PR flags as deliberate:

| Scenario | Notice shown |
|---|---|
| `simplefilter("ignore")` **before** `import aixplain` | **yes** — overridden (the documented sharp edge) |
| `simplefilter("ignore")` **after** `import aixplain` | no |
| `PYTHONWARNINGS=ignore` | no |
| `AIXPLAIN_SUPPRESS_V1_DEPRECATION=1` | no |

Row 1 does override an explicit user instruction. Given a sunset date that already passed unseen, I think it's the right trade and it's honestly documented and tested — but it's the team's call, not the PR's.

**2. Worth a release-note line:** anyone running `-W error::DeprecationWarning` (common in CI) will find v1 imports **hard-fail** after upgrading, since the warning is raised from inside the meta-path finder. That's inherent to the mechanism and shown as `exit 1` in the acceptance table, but downstream users will experience it as "the new version broke my build" unless it's called out where they'll read it.

**3.** `MIGRATION.md` naming the eight factories with no v2 equivalent — and making the removal date contingent on closing them — is the most useful part of this PR. Deprecating capability that has nowhere to land is how a sunset date slips a second time.

## Verified

- **Re-ran the acceptance table independently** (fresh interpreters, `-W error::DeprecationWarning`), all four rows match the PR:

  | Command | Exit |
  |---|---|
  | `import aixplain.v1` | 1 |
  | `import aixplain.factories` | 1 |
  | `from aixplain import Aixplain` | 0 |
  | `import aixplain.v1` + `AIXPLAIN_SUPPRESS_V1_DEPRECATION=1` | 0 |

- **The claim that matters most for v2 users holds:** after `import aixplain`, `[m for m in sys.modules if m.startswith("aixplain.v1")]` is **empty** — 0 v1 modules loaded, so v2 users never see the notice. `import aixplain` under `-W error::DeprecationWarning` exits 0.
- **The computed `stacklevel` actually works.** From a scratch app whose line 5 is `import aixplain.factories`, the notice is attributed to `user_app.py line 5` — the user's own import, not `importlib._bootstrap` and not `sys:1`. This is the hardest part of the PR (warnings from a `MetaPathFinder` sit under a variable 4–7 bootstrap frames) and the frame walk, which skips bootstrap frames the way CPython's own `is_internal_frame` does, gets it right.
- **Tests:** `pytest tests/unit` → **1600 passed** (merge-base `development` @ `42ed1e02`: **1490 passed**). No regressions; +110 new.
- **Warn-once reasoning checked against the code:** `create_module`/`load_module` both call it, keyed on a single `"v1"` sentinel — correct, because one `import aixplain.modules` fans out to ~160 redirected imports and v1 imports itself through the legacy paths (`aixplain/v1/enums/function.py` → `from aixplain.modules.model import Model`). Per-module keying would emit dozens of notices, which is what drives blanket suppression.
- `aixplain/v1/__init__.py` warning directly is necessary, not redundant: `import aixplain.v1.factories` bypasses a parent package's `__getattr__`, so a PEP 562 hook would stay silent on the most common usage.
- `AixplainV1DeprecationWarning(DeprecationWarning)` gives both `-W error::DeprecationWarning` matching and a dedicated category — verified by rows 1 and 2 of the table.
- `V1_REMOVAL_DATE` as single source of truth with `test_v1_deprecation_docs.py` failing on any doc that disagrees is the right shape: the constant changes first, docs follow.

**Not verified:** the doc prose itself — I did not read all 356 lines of `MIGRATION.md` line by line, only its structure and the "no v2 equivalent yet" list; the accuracy of each individual v1→v2 code mapping is best checked by whoever owns those factories.

PR #1045, opened 2026-08-12
fix(v2): prevent duplicate run submissions — POST retry, poll-retry, failed-classify (BUG-1090)

aix-ahmet:
## Summary

A single agent/model run could be submitted — and therefore **billed** — more than once through three independent paths. This PR closes all three (BUG-1090).

The unifying principle: a run submission is *billable and not idempotent*, so it may only be retried by a layer that knows what one costs. Transport-level machinery doesn't, so it no longer retries POST at all; the SDK does, so it owns the retry via an explicit, narrowly-scoped boundary.

### 1. Transport-level POST retry (`client.py`)

`create_retry_session` listed `POST` in urllib3's `allowed_methods`. Once a default read timeout existed, a model slower than `DEFAULT_TIMEOUT_READ` raised `ReadTimeout` *inside* urllib3, which then silently re-POSTed below the SDK — turning one `/execute` into up to `total + 1` submissions, invisibly.

Retries are now restricted to `GET` via the new `RETRY_ALLOWED_METHODS`.

Scope note, since urllib3 consults `allowed_methods` for only some retry classes: `status_forcelist` responses and *read* errors (`ReadTimeoutError`, `ProtocolError`) are method-gated, so POST gets neither. *Connect* errors (`ConnectTimeoutError`, `NewConnectionError`) are still retried for every method — urllib3 skips the method check there by design, and safely so: the connection was never established, so no request bytes reached the backend and no run was submitted.

### 2. Poll-time retry re-submitted a running job (`resource.py`)

`RunnableResourceMixin.run()` wrapped **both** the POST and the polling loop in its retry loop. A network blip while watching an already-running job therefore fell into the `except APIError` handler and restarted the entire run from the POST.

The retry boundary is now `_submit_with_retries`, which wraps the POST **alone**. Polling sits outside it and raises instead of starting a second execution.

### 3. Business `FAILED` responses classified as retryable (`exceptions.py`)

`create_operation_failed_error` built an `APIError` from a `FAILED` response body whose `statusCode` is usually absent — collapsing onto `0`, the exact sentinel `_is_retryable_run_error` reads as "no HTTP response at all, i.e. transport failure". A deterministic business failure was thus indistinguishable from a dropped connection, and got re-POSTed and billed again.

`APIError` gains a **tri-state** `retryable` flag:
- `None` (default) — no opinion; fall through to the existing status-code heuristic. This is what keeps all ~20 existing construction sites at their current classification.
- `False` — deterministic; re-submitting cannot fix it.

`create_operation_failed_error` now sets `retryable=False`, and `_is_retryable_run_error` honours an explicit flag over the heuristic.

## Incidental changes

- `Model._run_sync_v2` and `RunnableResourceMixin.run_async` each carried their own copy of the retry loop; both now delegate to `_submit_with_retries`. `model.py` drops its now-unused `time` / `APIError` imports.
- The added stack frame shifts the `max_iterations` deprecation warning `stacklevel` from 5 to 6 in `agent.py`, so it still points at the user's `agent.run(...)` call site rather than at SDK internals. `_submit_with_retries` sits on both the `run` and `run_async` paths, so both stay at the same depth.
- Docstrings for `run_retries` / `run_async` / `run` now state the guarantee explicitly: retries cover the submission only.

## Behavioural impact

`run_retries` semantics narrow deliberately — it retries *submissions*, not whole runs. A failure while polling, and an immediate business `FAILED`, now raise on the first occurrence instead of silently re-running. That is the fix, not a regression: the previous behaviour double-billed.

## Testing

363 unit tests pass across `test_resource.py`, `test_client.py`, `test_exceptions.py`, `test_model.py`, and `test_agent_budget.py`, including new coverage for each of the three paths. Full pre-commit suite (ruff, ruff format, pytest) passes.

Note: this PR targets `bug-937-v2-validate-poll-url-host-and-dr`, not `main`.


elbadrashiny (approved):
**Verdict:** approve — all three defects are real, the fixes are correct, and the two claims most likely to be wrong (urllib3's method gating and the `stacklevel` shift) both check out against source and at runtime. Two non-blocking notes about scope.

## Non-blocking

**1. Dropping POST from `allowed_methods` also removes 5xx retry from POSTs that aren't billable.** The rationale — "a submission is billable and not idempotent" — applies to `/execute`. But `create_retry_session` is shared, so the same change silently removes transparent 500/502/503/504 retry from:

- `upload_utils.py:162` — `POST sdk/file/upload/temp-url` (presigned-URL request). Not billable, safely retryable, and it now fails on a transient backend 503 that previously recovered.
- `issue.py:64` — `client.post(self._issue_url(), …)`.

Worth confirming this is intended rather than collateral. If not, the narrow fix is a separate session for the upload path (it already builds its own via `RequestManager.create_session`) or a per-request retry override. Note this becomes live only once #1041 lands too, since that's what puts a default read timeout on the upload legs.

**2. `create_operation_failed_error` now marks *every* `FAILED` body non-retryable.** The motivating case is the absent `statusCode` collapsing onto `0`, but the flag also overrides the heuristic for a `FAILED` body that carries a genuine `statusCode: 503`. The docstring says "always marked non-retryable", so this reads as deliberate — just flagging that the narrowing is a bit wider than the stated motivation.

## Verified

- **Tests:** `pytest tests/unit` → **1623 passed, 4 skipped**, against its base (`pr-1040`, **1604 passed**). No regressions; +19 new.
- **urllib3's method gating, read from the installed source (2.7.0)** rather than assumed — the scope note in the code comment is exactly right:
  - `Retry.is_retry()` opens with `if not self._is_method_retryable(method): return False` → `status_forcelist` **is** method-gated.
  - `Retry.increment()` read-error branch: `if read is False or method is None or not self._is_method_retryable(method): raise reraise(...)` → read errors **are** method-gated.
  - connect-error branch decrements `connect` with **no** method check → still retried for POST, and safely, since no bytes reached the backend.
- **Mutation-tested all four load-bearing changes** — each reverted individually, all go red:
  - `POST` back in `RETRY_ALLOWED_METHODS` → 4 failed
  - `retryable=False` dropped from `create_operation_failed_error` → 5 failed
  - explicit-flag branch removed from `_is_retryable_run_error` → 3 failed
  - `stacklevel` 6 → 5 → 3 failed
  - polling moved back inside the retry boundary (`_submit_with_retries` → `_post_and_handle_run`) → 3 failed
- **`stacklevel=6` is correct for both paths, and pinned.** `Agent` defines *both* `run` (agent.py:1019) and `run_async` (agent.py:1054), so the two chains to `build_run_payload` are the same depth — the claim holds structurally, and `test_agent_budget.py` asserts `w.filename == __file__` for each, which is a real discriminator (it fails at 5).
- **Retry boundary read directly:** `run()` now calls `_submit_with_retries` and does `sync_poll` outside it; `run_async` and `Model._run_sync_v2` both delegate to the same helper, so the three copies of the loop collapsed to one with no behavioural divergence between them.
- `model.py`'s dropped `time` / `APIError` imports have no remaining references.

**Not verified:** functional suites (need live credentials); no end-to-end proof against a real slow model that a `ReadTimeout` no longer produces a second submission — that needs the backend. The unit-level proof is a mocked transport, which is why I mutation-tested instead of trusting it.

PR #1046, opened 2026-08-12
ENG-3544: make agent/team_agent functional skips conditional; re-attach orphaned CI legs

aix-ahmet:
## Problem

The `agent` and `team_agent` functional CI legs were reporting **green while executing 0 of 50 collected tests**.

Each directory's `conftest.py` added an unconditional skip marker in `pytest_collection_modifyitems`. Since pytest exits `0` for an all-skipped session and the workflow has no other signal, a leg that skips everything is indistinguishable from a leg that passed. Measured on the defect state: `agent` collects **35** items / executes **0**; `team_agent` collects **15** / executes **0** — roughly 2,300 LOC of coverage that silently did nothing.

Separately, commit `04dea96e` removed four leg names from the CI matrix but kept their test files, orphaning `tests/functional/benchmark`, `pipelines/designer_test.py`, and `pipelines/create_test.py` as dead code that no job ever ran.

## Changes

**1. Conditional skips instead of blanket skips** (`tests/functional/{agent,team_agent}/conftest.py`)

Both directory-wide skips are replaced with an autouse, function-scoped fixture that skips only when neither `TEAM_API_KEY` nor `AIXPLAIN_API_KEY` is set — the pattern already used by `tests/functional/v2/conftest.py`. Applied at setup time rather than at collection, so `--collect-only` still reports the true inventory. Comments in both files record why the directory-wide form must not come back: if an individual test is failing, skip or xfail *that test* with a ticket in the reason.

**2. Execution guard** (`tests/ci_guards.py`, wired in `tests/conftest.py`)

New `pytest_sessionstart` / `pytest_runtest_logreport` / `pytest_sessionfinish` hooks fail any session that reports success (exit `0` or `5`) while running no test body. This is the directory-scale counterpart to the existing per-test assertion guard.

Design points, each covered by a test:
- **Opt-in** via `AIXPLAIN_REQUIRE_EXECUTED_TESTS`, which the workflow now sets. Local `pytest -k ...`, single-file runs, and no-credential laptops behave exactly as before.
- Only `call`-phase reports count — a plain skip never reaches `call`, whether it comes from a marker or from `pytest.skip()` in a fixture.
- **xfails count as executed** (the body did run); `wasxfail` is the discriminator.
- **xdist workers are exempt** — a worker handed nothing but skips would otherwise fail a run that executed plenty elsewhere. The controller sees every worker's reports, so it holds the true total.
- **`--collect-only` and friends are exempt** (`--fixtures`, `--setup-only`, `--setup-plan`). Not hypothetical: the workflow exports the flag for every job and `test_credential_free_collection.py` spawns `pytest tests/unit --collect-only` with an inherited environment.
- **Already-failing sessions are left alone**, so a real error is never rewritten as "executed nothing".
- The ledger resets per session, since it outlives a single session when pytest runs in-process.

Helpers live in `tests/ci_guards.py` (a name pytest does not collect) so they can be unit-tested directly; the conftest only wires them into hooks.

**3. Re-attached orphaned legs** (`.github/workflows/main.yaml`)

`benchmark`, `pipeline_designer`, and `pipeline_create` are back in the matrix. The two pipeline legs deliberately pass no `--sdk_version` / `--pipeline_version` flags: the filters in `tests/conftest.py` drop every item lacking the named parameter, so passing them to a non-parametrised file would collect N and execute 0 — the exact failure mode this PR removes.

**4. Matrix invariants as unit tests** (`tests/unit/test_ci_matrix_coverage.py`)

An orphaned test file is now a unit-test failure rather than silent dead code. The suite asserts that every functional directory is claimed by exactly one leg, every leg path exists and declares a timeout, no functional conftest blanket-skips its directory, and the workflow enables the execution guard. The blanket-skip detector distinguishes a guarded conftest from an unconditional one (a bypass in an earlier version of this detector was caught in review and fixed).

## Verification

- `pytest tests/unit/test_ci_guards.py tests/unit/test_ci_guard_wiring.py tests/unit/test_ci_matrix_coverage.py` → **54 passed**. The wiring tests drive real pytest sessions through the `pytester` fixture rather than asserting on mocks.
- Regression check against the original defect: a session in the old state (35 collected, 0 executed) exits **1** under the guard; the fixed tree exits **0** locally without the flag set.
- Full pre-commit suite passes on the commit.


elbadrashiny (changes requested):
**Verdict:** request changes — the guard itself is well built and I verified it works, but the PR re-enables 50 tests and 3 CI legs that were deliberately disabled three weeks ago *because they were failing*, without establishing the precondition its own deleted comment named. Two concrete items.

## Blocking

**1. This reverses a deliberate decision, and the "once they're green" precondition isn't shown to be met.**

The PR frames both blanket skips as an accident — "a leg that skips everything is indistinguishable from a leg that passed". But `git log` says they were intentional:

```
04dea96e  2026-07-23  Hadi Nasrallah
test: prune flaky/failing functional tests and trim CI matrix (#1003)

 .github/workflows/main.yaml              | 16 --
 tests/functional/agent/conftest.py       | 14 ++
 tests/functional/team_agent/conftest.py  | 14 ++
 tests/functional/model/run_utility_model_test.py | 265 ---
 tests/functional/pipelines/designer_test.py      |  39 +-
```

That one commit **created** both conftest skips, removed the matrix legs, and deleted 265 lines of flaky tests. The comment this PR deletes stated the precondition explicitly:

> `# Skip the entire agent functional test suite. Re-enable by removing this file once the agent tests are green against the test backend.`

The same applies to `benchmark`, `pipeline_designer` and `pipeline_create`: the PR describes them as "orphaned as dead code", but `04dea96e`'s subject is *"trim CI matrix"* — removing those legs was the point, not an oversight.

The verification section covers unit tests (54 passed) and a synthetic replay of the defect state. It does not claim the 50 functional tests or the 3 re-attached legs pass against the test backend. If they don't, merging turns five CI legs red — and #1048, stacked above this, is about making contexts *required*.

**What would resolve it:** a run showing `agent`, `team_agent`, `benchmark`, `pipeline_designer` and `pipeline_create` green against the test backend. If some aren't ready, land the guard and the conditional-skip change (both good) and keep the not-ready tests parked with per-test `skip`/`xfail` carrying a ticket — which is exactly the discipline the new conftest comments ask for.

**2. The three re-attached directories have no credential guard — unlike the two this PR fixes.**

`agent` and `team_agent` get an autouse `_require_api_key` fixture. `tests/functional/benchmark`, `pipelines/designer_test.py` and `pipelines/create_test.py` get nothing, so without a key they don't skip, they fail against the live backend:

```
tests/functional/benchmark                    -> EXIT=1 | 4 failed
tests/functional/pipelines/designer_test.py   -> EXIT=1 | 6 failed, 6 errors
tests/functional/pipelines/create_test.py     -> EXIT=1 | 3 failed, 1 passed
```

The failures are `403 Forbidden` from `platform-api.aixplain.com` and `Pipeline Deletion Error: Make sure the pipeline exists and you are the owner` — i.e. they reached production unauthenticated. The same autouse fixture applied to these three would make them consistent with the rest of `tests/functional`.

## Non-blocking

`test_ci_matrix_coverage.py` asserting that every directory under `tests/functional/` is claimed by exactly one matrix leg is a **policy change**, not just an invariant: it removes "park a flaky directory" as an option — the team would have to delete the files instead. Given `#1003` shows parking is a tool they use, that deserves an explicit yes from whoever owns CI.

## Verified

- **Tests:** `pytest tests/unit` → **1582 passed** (base `pr-1043`: 1528). No regressions.
- **The 50-test claim:** `pytest tests/functional/agent tests/functional/team_agent --collect-only` → **50 tests collected**. Matches.
- **The guard works, at the level that matters — the process exit code:**
  - guard on, no credential, all 35 agent tests skip → **exit 1**, with the `CI integrity failure (ENG-3544): 35 test(s) collected, 0 executed` banner
  - guard off (control) → **exit 0**
  - guard on with `--collect-only` (the exemption that keeps `test_credential_free_collection` working in CI) → **exit 0**
- **Read the implementation:** only `call`-phase reports count so a fixture-raised skip never counts; `wasxfail` makes xfails count as executed; the ledger resets per session because it outlives one when pytest runs in-process; xdist workers exempt via `workerinput`; already-failing sessions left alone so a real error isn't rewritten. `session.exitstatus = 1` in `pytest_sessionfinish` does take effect — confirmed by the exit codes above, not assumed.
- **The `pipeline_designer` / `pipeline_create` flag omission is correctly reasoned:** `filter_items` in `tests/conftest.py` drops every item lacking the named param, and it only runs when the CLI option is passed — so passing `--sdk_version` to a non-parametrised file would indeed collect N and execute 0.

**Not verified:** whether the re-enabled tests pass *with* a credential against the test backend — that is item 1, and it needs a credential I shouldn't use; the workflow YAML changes as executed by GitHub Actions.

elbadrashiny (approved):
**Verdict:** approve — the credential-guard gap is closed, and the parking response is better than what I asked for. One point stays open, but it's a CI-strategy call rather than a defect, so I'm not blocking on it.

## Verified on `4f8b476`

**The credential guard is applied consistently.** `tests/functional/benchmark/conftest.py` and `tests/functional/pipelines/conftest.py` now carry the same autouse, function-scoped fixture as `agent`/`team_agent`, so a keyless run skips instead of failing on unauthenticated 403s against the live backend.

**The parking response is the better answer.** Rather than just reverting the three legs, `PARKED_TARGETS` puts the backlog in the repo with a ticket reference per path and cites `04dea96e` as the reason — and the coverage audit now accepts *a leg or a declared parking entry*, so silent orphaning is still a failure. That's a genuine improvement on the status quo in both directions: nothing can be quietly parked, and nothing can be quietly orphaned.

**All three parking guards fire** (I mutated each rather than trusting them):
- dropping the `benchmark` entry while its files remain → fails `test_every_functional_directory_maps_to_a_leg_or_is_fully_parked`
- a parking entry with no ticket reference → fails `test_parked_targets_cite_a_ticket[tests/functional/benchmark]`
- re-adding `benchmark` as an active leg while it's still parked → fails `test_matrix_names_and_include_entries_agree`

`pytest tests/unit` → **1593 passed** with no credential. The execution guard's behaviour is unchanged from my first pass, where I confirmed it exits 1 on an all-skipped session, 0 when off, and 0 under `--collect-only`.

## Open, non-blocking

`agent` and `team_agent` remain **active matrix legs**, and this commit doesn't touch their conftests — so the 50 tests `04dea96e` blanket-skipped as flaky/failing will execute in CI for the first time, still without evidence they pass against the test backend. By this PR's own new standard — park what's unproven, with a ticket — they'd arguably belong in `PARKED_TARGETS` alongside the other three.

I'm not blocking on it, for two reasons: leaving them active is the ticket's actual deliverable, and "red and real" beats "green and fake" — especially since `main.yaml` is push-only and `required_status_checks.contexts` is still empty, so a red leg there blocks nobody today. It's also not something I can settle by reading; it needs a run with a test-backend credential.

So, purely as a heads-up for whoever merges: expect those two legs to be the first honest signal anyone has had from them in three weeks, and be ready for that signal to be red. If it is, the fix is a `PARKED_TARGETS` entry each rather than a revert.

**Not verified:** whether the re-enabled `agent`/`team_agent` tests pass with a credential (the point above); the workflow changes as executed by GitHub Actions.

PR #1047, opened 2026-08-12
ENG-3543: fix package discovery glob so all modules ship; add build-import CI check

aix-ahmet:
## The bug

`[tool.setuptools.packages.find].include` patterns are fnmatch'd against the **full dotted package name**, so `include = ["aixplain", "tests"]` matched only the top-level package and silently discarded all 26 subpackages.

Two unrelated accidents hid this. `setuptools-scm` was in `build-system.requires` with no `[tool.setuptools_scm]` section anywhere in the repo — its only effect was registering a git `file_finders` sweep that pulled every tracked file into `SOURCES.txt`. And a stale `aiXplain.egg-info/SOURCES.txt` rode along inside the sdist.

Remove both crutches (build from a `.git`-less export, as any sdist consumer or CI checkout without full history does) and the result is:

- **3 of 169 modules shipped** — no error, no warning
- the wheel **installs cleanly**
- first import raises `ModuleNotFoundError: No module named 'aixplain.v2'`

## The fix

**`pyproject.toml`**
- `include = ["aixplain*", "tests*"]` — the trailing glob is the actual fix
- `exclude = ["*.__pycache__", "*.__pycache__.*"]`
- dropped `setuptools-scm` from build-requires; pinned `setuptools>=61` (the floor for `[project]`-table metadata)
- documented `namespaces = true` as load-bearing: `aixplain/v1/base/` and `aixplain/v1/factories/cli/` hold modules but no `__init__.py`, and are reached through the `aixplain._compat` redirector (`aixplain.base.parameters`, `aixplain.factories.cli.model_factory_cli`)

**`setup.cfg`**
- `license_files=LICENSE`. The old `LICENSE.rst` pattern matched no file (`Pattern 'LICENSE.rst' did not match any files`) and only "worked" because the git sweep hauled `LICENSE` into the sdist regardless. Dropping the sweep without this would have shipped an Apache-2.0 distribution with **no licence file at all** — a mismatched pattern is only a deprecation warning, so nothing else catches it.

## The guard — two halves, neither subsumes the other

**`package-integrity` job (`.github/workflows/main.yaml`)** — the slow half. A build that succeeds only because `.git` happens to be present is not a build, so the job removes both crutches:

1. `rsync` the tree excluding `.git`, `*.egg-info`, `__pycache__`, `dist`, `build`, then assert both crutches are actually gone (either one alone still masks the bug)
2. `python -m build` with no `--wheel`, so the flow is sdist → wheel *from that sdist* — which also proves the sdist is self-sufficient
3. assert every `.py` under `aixplain/` in the source tree is present in **both** artifacts. The expected set is derived from the tree, never hardcoded — adding a module must not require editing the gate, and a hardcoded count rots on the first PR. Includes a canary check that the two `__init__`-less dirs are still in the scan, and a check that the sdist carries `LICENSE`
4. install the wheel into a clean venv and import 11 canary modules — one per top-level subpackage plus both legacy-redirect paths. `cd` out of the repo **first**, since a cwd-local `aixplain/` shadows site-packages and would make this step pass on a 3-module wheel. Deliberately not a `walk_packages()` sweep: modules behind optional extras (plotly, docker, model-interfaces) would fail for unrelated reasons and make the gate flaky
5. `aixplain --help` to confirm the console script is wired up

Credential-free on purpose: if it fails on a missing key, the regression is import-time credential validation, not this workflow.

**`tests/unit/test_packaging_config.py`** — the fast half, run on every branch push via pre-commit (the CI job only runs on main/test). Six tests: the include patterns cover every real package dir under `aixplain/`; `exclude` discards nothing real; `namespaces` stays true while those dirs lack `__init__.py`; `setuptools-scm` has not returned unconfigured; `license_files` matches a real file; and the CI job still performs its three load-bearing steps (name-only would pass on a gutted job).

The static test cannot see a setuptools upgrade changing discovery semantics; the CI job does not run on feature branches. Hence both.

**`.pre-commit-config.yaml`** — added `tomli>=2.0.1; python_version < '3.11'` (also to the `test` extra). The hook builds its own isolated env and never sees the extra, so the dep is repeated in both places.

## Verification

Built and installed the package rather than reading the diff alone. From a `.git`-less, egg-info-less export:

```
source=169  wheel=169  sdist=169   PROBLEMS: none
11 canaries OK
```

Before the fix, the same procedure yielded 3 of 169. Locally: `6 passed` for the new unit tests, and the full pre-commit suite (including the pytest gate) passes on the commit.


elbadrashiny (changes requested):
**Verdict:** request changes — one line. The core fix is real, important, and I proved it end to end; but the same widened glob now ships the entire test suite into users' `site-packages`, and that is much cheaper to fix before a release than after.

I reproduced the bug and the fix from a `.git`-less, egg-info-less export, exactly as the PR prescribes:

| | source | wheel | sdist | LICENSE in sdist |
|---|---|---|---|---|
| base (`pr-1046`) | 169 | **3** | **3** | **0** |
| this PR | 169 | **169** | **169** | **1** |

The diagnosis is precise and the `setup.cfg` catch is a good one — dropping the `setuptools-scm` git sweep without fixing `license_files` would have shipped an Apache-2.0 package with no licence file, and a mismatched `license_files` pattern is only a deprecation warning.

## Blocking (one line)

`include = ["aixplain*", "tests*"]` — the `tests*` glob takes the shipped test suite from **5 files to 117**:

```
base wheel:  3 aixplain/*.py     5 tests/*.py
this PR:   169 aixplain/*.py   117 tests/*.py
```

and after installing the wheel into a clean venv:

```
>>> import tests
YES -> /private/tmp/cleanvenv/lib/python3.13/site-packages/tests
```

So `pip install aiXplain` puts an importable top-level `tests` package into `site-packages` — including `tests/functional/*`, which hits the live backend. A top-level `tests` in site-packages can shadow (or be shadowed by) a user's own `tests` package depending on `sys.path` order, and wheel size goes 17K → 878K.

The 5-file version of this already existed, so it isn't new — but this PR is specifically about *what ships*, it multiplies the surface 23×, and #1048 (stacked directly above) is what makes it publishable. PyPI filenames are immutable; the PR makes that point itself.

**Fix:** `include = ["aixplain*"]`, and update `test_every_package_dir_is_matched_by_an_include_pattern` in `tests/unit/test_packaging_config.py`, which currently asserts the `tests` dirs *are* matched. Nothing needs `tests` to be installed — `tests/conftest.py`'s `from tests.ci_guards import …` resolves via pytest's rootdir insertion when running from the repo, not from site-packages.

If shipping tests is deliberate (some consumer runs them post-install?), say so in the comment and I'll withdraw this.

## Verified

- **Built the package rather than reading the diff.** `rsync` excluding `.git`/`*.egg-info`/`__pycache__`/`dist`/`build`, asserted both crutches gone, then `python -m build` with no flags (sdist → wheel *from that sdist*). Numbers in the table above.
- **Installed the resulting wheel into a clean venv and imported from outside the repo** (so a cwd-local `aixplain/` couldn't mask a broken wheel): `aixplain`, `aixplain.v1`, `aixplain.v2`, `aixplain.utils.config`, `from aixplain import Aixplain`, plus both `__init__`-less legacy redirect paths — `aixplain.base.parameters` and `aixplain.factories.cli.model_factory_cli`. All resolved to `site-packages`. `aixplain --help` works, so the console script is wired.
- **`namespaces = true` really is load-bearing** — those last two canaries are the proof, and they fail without it.
- **Tests:** `pytest tests/unit` → **1588 passed** (base `pr-1046`: 1582). No regressions; +6 new.
- **Stacking order is right, and #1048's guard would have caught this bug:** a GitHub checkout still has `.git`, so without this PR the `setuptools-scm` sweep would populate the sdist, and the wheel built *from that sdist* would drop to 3 modules — which #1048's clean-venv canary import step catches. Good defence in depth, in the correct order.

**Not verified:** the `package-integrity` CI job as executed by GitHub Actions (I ran its logic by hand, above); behaviour under a future setuptools that changes discovery semantics — which is exactly why the PR keeps both the static test and the CI job.

elbadrashiny (approved):
**Verdict:** approve — the one blocking item is fixed, and I re-verified it by rebuilding rather than reading the diff.

## Verified on `e47add7`

- `include = ["aixplain*"]`; the `tests*` pattern is gone.
- **Rebuilt from a `.git`-less, egg-info-less export** (the same procedure as my first pass): `source=169 wheel=169 sdist=169`, **0 test files in the wheel** (was 117), and `LICENSE` still present in the sdist. So the packaging fix is intact and the test suite no longer lands in users' `site-packages`.
- The new guard is not decorative: re-adding `tests*` to `include` fails `test_the_test_suite_is_not_packaged`.
- `pytest tests/unit` → **1589 passed** with no credential.

Thanks for taking the `from tests.ci_guards import …` point — you're right that it resolves through pytest's rootdir insertion, so nothing needed the suite installed.

**Not verified (unchanged from my first review):** the `package-integrity` CI job as executed by GitHub Actions; behaviour under a future setuptools that changes discovery semantics.

PR #1048, opened 2026-08-12
ENG-3428: add release pipeline and least-privilege workflow permissions; document required CI contexts on main

aix-ahmet:
## Why

0.2.47 was uploaded to PyPI **by hand from a developer laptop**. Nothing ties that file to a commit, and a long-lived API token is the only credential guarding the project name.

Separately, `main`'s `required_status_checks.contexts` is **empty** and `GET /rulesets` returns `[]` — so a PR with red CI merges (PR #999 merged with 8 failing checks). Every gate the surrounding ENG-34xx chain adds — the 62% coverage floor, the `.git`-less package-integrity build, the zero-executed-test guard — is **advisory** until at least one context is marked required.

### `.github/workflows/release.yaml` (new)

Tag-push-only (`v*`) — the published artifact must be built from an immutable ref, and the tag is the version's only identity. Deny-all `permissions: {}` at workflow level; every scope granted per job, so a job added later starts with nothing.

**`build`** — `contents: read`, deliberately **no** `id-token`:
- Asserts the tag matches `[project].version` by **PEP 440 version equality**, not string equality (`v0.2.48.rc1` and `0.2.48rc1` normalise to one PyPI filename). PyPI filenames are immutable — a mislabelled upload can only be yanked.
- Runs the static packaging guards with `AIXPLAIN_REQUIRE_EXECUTED_TESTS=1`, so a collection-time skip can't exit 0 and let the release proceed on an unchecked tree.
- `python -m build` with no flags: sdist → wheel *from that sdist*, which also proves the sdist is self-sufficient. Wipes `*.egg-info` first — a stale `SOURCES.txt` riding inside the sdist was one of the two accidents that hid ENG-3543.
- Installs the wheel into a clean venv, `cd`s **out of the repo** (a cwd-local `aixplain/` would shadow site-packages), imports across v1/v2, and runs `aixplain --help`. The ENG-3543 wheel shipped 3 of 169 modules and installed cleanly — "pip install succeeded" proves nothing on its own.

**`publish`** — `id-token: write` and nothing else, `environment: pypi`, **no checkout, no pip, no repository code**. Downloads the artifact and calls `pypa/gh-action-pypi-publish` with attestations. A compromised build dependency must never share a process with the OIDC credential that can upload to PyPI.

### `.github/workflows/{main,pre-commit}.yaml`

Add `permissions: contents: read`. Both execute repository code and third-party dependencies (pre-commit runs hook code from pinned third-party repos) while inheriting the repository default token, which is **write-all** here. `docs.yaml` already declared its own scopes, so all four workflows are now explicit.

### `docs/ci/required-checks.md` (new)

> **Describes repository *settings* changes. Nothing here is applied by merging this PR.**

Requiring contexts while CI is red would block every merge — including the PRs that make CI green. So the workflows ship now and the settings flip is a documented one-time follow-up.

- Contexts in two tiers, with the matrix-naming gotcha (`include`-only keys aren't part of the check name, so it's `setup-and-test (agent)`, not `setup-and-test (agent, tests/functional/agent, 45)`).
- Why `main.yaml` contexts **cannot** be required today: it triggers only on `push` to `main`/`test`, so a feature-branch PR never produces them — requiring them now is a merge deadlock, not a merge gate. `pre-commit` is the only context that reports on a PR today.
- Ordered procedure, the `gh api` invocation (with the `-F` vs `-f` trap: `-f strict=true` sends the string `"true"` and 422s), the fork-PR caveat and why `pull_request_target` is not an acceptable substitute.
- One-time release setup: `pypi` environment **with a `v*` tag rule** (nothing in `release.yaml` can enforce which ref a tag points at — that restriction can only live in the environment protection rule), PyPI trusted-publisher registration, a TestPyPI dry run, token revocation, and flipping default workflow permissions to read-only.
- Recorded follow-ups (SHA-pin the publish action; extract `package-integrity` into a `workflow_call` reusable workflow; drop `twine`/`pre-commit` from runtime deps) and four open questions.

### `tests/unit/test_ci_permissions.py` (new)

Static assertions over the workflow files: every workflow declares `permissions`, no job exceeds what it needs, `publish` holds `id-token` and nothing else and takes no `password:`/`user:` input, and **the Tier 2 context table in the doc matches `main.yaml`'s matrix exactly** — so renaming a matrix leg fails a unit test instead of silently leaving a stale required-context list in the doc.

## Verification

- `pytest tests/unit/test_ci_permissions.py tests/unit/test_packaging_config.py` — 26 passed.
- Full pre-commit suite passed on commit.

## Note for reviewers

`release.yaml` is **inert** until the manual setup is done, and merging it is safe in the meantime: `build` passes, and `publish` fails at the OIDC exchange with a clear PyPI error. Nothing is published and nothing else breaks.

Based on `eng-3543-fix-package-discovery-glob-so-a`.


elbadrashiny (changes requested):
**Verdict:** request changes — the workflow design is genuinely good (deny-all default, per-job scopes, an OIDC job that touches no repository code), but the single step holding the PyPI upload credential is pinned to a mutable branch, and the release gate runs two test files. Both are one-line fixes and both belong in the PR that creates the publish path.

## Blocking

**1. `pypa/gh-action-pypi-publish@release/v1` is a mutable ref on the one step that can publish to PyPI.**

```yaml
- name: Publish to PyPI with attestations
  uses: pypa/gh-action-pypi-publish@release/v1   # <- branch, not a SHA
```

Everything else about `publish` is exemplary — no checkout, no pip, `id-token: write` and nothing else — and then it hands that token to code fetched from a branch that can change under you. Anyone who can move `release/v1` (upstream compromise, maintainer account takeover) publishes arbitrary code as aiXplain. The comment records SHA-pinning as a follow-up in `docs/ci/required-checks.md`; for the step that holds the upload credential, I'd do it here:

```yaml
uses: pypa/gh-action-pypi-publish@<full-40-char-sha>  # release/v1 == vX.Y.Z
```

The rest of the repo isn't SHA-pinned either, so this isn't a regression — but this PR is what creates the privileged step, and `test_ci_permissions.py` is the natural place to assert the pin holds.

**2. A tag can publish to PyPI without the unit suite ever having run.**

The `build` job's only test step is:

```yaml
python -m pytest tests/unit/test_packaging_config.py tests/unit/test_ci_permissions.py -q
```

The comment says the tagged commit "must already have passed" `main.yaml`'s `package-integrity`, but nothing enforces that — `main.yaml` triggers only on push to `main`/`test`, and `required_status_checks.contexts` is empty, which this PR documents itself. So `git tag v0.2.48 <any-commit> && git push --tags` publishes to PyPI with the unit suite unrun.

Adding `python -m pytest tests/unit` to the `build` job closes it, and it's free now: #1043 made the suite credential-free, and I ran it that way (1608 passed, no key). Given the release job already installs `.[test]`, this is one line.

## Non-blocking

- `docs/ci/required-checks.md` is unusually good — the `include`-only matrix-naming gotcha, `-F` vs `-f` sending `"true"` as a string, and *why* `main.yaml`'s contexts can't be required today (push-only trigger → merge deadlock, not a merge gate) are all things that would otherwise be learned the expensive way. The "nothing here is applied by merging this PR" banner is the right call.
- Tying the `pypi` environment to a `v*` tag rule is load-bearing and can only live in repo settings, as the doc says. Worth calling out in the merge checklist so it isn't missed — until it exists, tag-push protection is the only thing standing between a pushed tag and PyPI.
- PEP 440 version *equality* rather than string comparison (`v0.2.48.rc1` vs `0.2.48rc1`) is the right call and the kind of thing normally got wrong.

## Verified

- **Tests:** `pytest tests/unit` → **1608 passed** with no credential (base `pr-1047`: 1588). No regressions; +20 new.
- **Read every workflow permission block.** `permissions: {}` at workflow level; `build` has `contents: read` and no `id-token`; `publish` has `id-token: write` only, `environment: pypi`, no `checkout`, no `pip`, no repository code — the build/publish privilege separation the description claims is actually implemented.
- **Enumerated every `uses:` across all four workflows** — item 1 above is the only one holding a credential; the rest are `actions/*@v4`, consistent with existing practice.
- Confirmed `main.yaml` and `pre-commit.yaml` gain `permissions: contents: read`, so all four workflows now declare scopes explicitly (`docs.yaml` already did).
- Confirmed the `publish` step passes no `password:`/`user:` input, so Trusted Publishing is the only path, and `test_ci_permissions.py` fails if a token input reappears.

**Not verified:** any of this as executed by GitHub Actions — I can't run workflows here, so the release path is reviewed statically. In particular the OIDC exchange, the `pypi` environment gate, and the claim that `publish` fails cleanly at the OIDC step before setup is done are all unexercised. That last one matters for the "merging is safe in the meantime" argument, and it's worth a TestPyPI dry run before the first real tag.

elbadrashiny (approved):
**Verdict:** approve — both blocking items are fixed, and both are now pinned by guards that I confirmed fire.

## Verified on `19f17ed`

**1. The publish action is SHA-pinned, and the pin is the right commit.** I didn't take the comment on trust — `v1.14.2` is an *annotated* tag, so I dereferenced it through the API:

```
repos/pypa/gh-action-pypi-publish/git/ref/tags/v1.14.2 -> tag object a892a5a6…
repos/pypa/gh-action-pypi-publish/git/tags/a892a5a6…   -> commit dc37677b2e1c63e2034f94d8a5b11f265b73ba33
```

which is exactly what `release.yaml:178` pins, and it's a full 40-hex SHA. The `# v1.14.2` comment is accurate.

**2. The release gate now runs the whole unit suite** — `python -m pytest tests/unit -q` in the `build` job, credential-free, so a tag can no longer publish a commit whose tests never ran.

**Both guards fire:**
- reverting the pin to `@release/v1` → fails `test_every_pypi_publish_step_is_pinned_to_a_commit_sha`
- deleting the unit-suite step → fails `test_release_build_job_runs_the_whole_unit_suite_before_publishing`

Asserting the ref must be 40 hex characters (rather than "not a branch name") is the right shape — it can't be satisfied by a tag either, which is what you want here.

`pytest tests/unit` → **1610 passed** with no credential.

**Still worth doing before the first real tag, and outside this PR:** the one-time repo setup in `docs/ci/required-checks.md` — the `pypi` environment *with its `v*` tag rule* (nothing in the workflow can restrict which ref a tag points at), trusted-publisher registration, a TestPyPI dry run, and revoking the existing long-lived token.

**Not verified:** anything as executed by GitHub Actions — the OIDC exchange, the `pypi` environment gate, and the claim that `publish` fails cleanly at the OIDC step before setup is done are all still static-only. That last one underwrites "merging is safe in the meantime", so the TestPyPI dry run is the thing that would close it.

PR #1055, opened 2026-08-14
feat(v2): add unified File assets and persistent agent files

xainaz:


elsheikhams99 (changes requested):
# Code Review — ENG-3524: unified `File` assets & persistent agent files

**Branch:** `ENG-3524-files-folders-one-aix-file-covering-files-and-directories-and-attaching-files-to-agents`
**Reviewed:** commits `5c963edb` + `8e23edff` against merge-base `f08c1c93`
**Reviewer:** Abdelrahman Elsheikh
**Test status:** all 1276 v2 unit tests pass — every blocking finding below is invisible to the current suite

Scope of the diff: new unified `aixplain/v2/file.py` `File` asset, persistent `Agent.files`, the `FileType`/`FileContentType` enum split, a client list-error fix, and two new unit test modules.

Top five findings were reproduced locally with scripted probes.

---

## Blocking — must fix before merge

### B1. Every local-file `save()` can crash on a read-only property
`aixplain/v2/file.py:107`

`File.__init__` blindly `setattr`s unknown backend keys. `_save_file` itself POSTs `url`, so a response echoing `url` raises `AttributeError: property 'url' of 'File' object has no setter`. This is the primary path for the feature. **Reproduced.**

### B2. `url` / `s3_url` / `relative_path` are never populated
`aixplain/v2/file.py:100`

The camelCase alias map omits `s3Url` and `relativePath`; values land on stray `s3Url`/`relativePath` attributes and `f.url` is `None` after a successful save. Breaks the documented SQLite flow in `skills/aixplain-builder/references/tools-integrations.md`, which passes `config={"url": resource.url}`. A `File` whose URL cannot be read is not usable downstream. **Reproduced.**

### B3. Second `save()` on the same object raises `ValidationError`
`aixplain/v2/file.py:188`

`_apply_data` copies every dataclass field from the response-hydrated object, nulling `source`/`file_path`, `description`, `size`, and `extension`. **Reproduced.**

### B4. User metadata silently dropped on file upload
`aixplain/v2/file.py:216`

`_save_file` posts only `{name, fileType, url}`, discarding user-supplied `description`/`tags`/`privacy`. Silent data loss, and inconsistent with the folder path, which does send them. **Reproduced payload.**

### B5. Wrong file kept / misleading error on list mutation
`aixplain/v2/agent.py:809`

`_sync_file_references` remaps `None` placeholders positionally:
- `del agent.files[0]` with two unsaved `File`s keeps the wrong file.
- `agent.files.insert(0, "id")` leaves `None` in `_original_files` → `ValueError("An agent file must be a File instance…")`.

Silent wrong-data plus an error that misdirects the user. **Both reproduced.**

### B6. Detaching all files from an agent is impossible
`aixplain/v2/agent.py:2043`

An empty list pops `"files"` from the payload, so clearing every file can never be persisted. Subset removal works, which makes the gap easy to miss. **Reproduced.**

---

## Non-blocking — should fix, not release-gating

### N1. `save(**_)` discards all kwargs
`aixplain/v2/file.py:275`

`save(is_temp=False)` (previously the permanent-upload endpoint, `aixplain/utils/file_utils.py:279`) and base-class `save(name=…/description=…)` shortcuts are silent no-ops; also bypasses `@with_hooks`. Borderline — promote to blocking if `is_temp=False` was a supported public call.

### N2. Enum split left half-done
`aixplain/v2/enums.py:20`

`FileType` lost its content members (moved to `FileContentType`) while `Resource` is still advertised as a compatible alias with re-purposed `file_type` semantics. `FileContentType` was never added to `enums_include.py` or the `Aixplain` class attrs. Ergonomic/consistency issue, not a runtime break.

### N3. Pre-upload validation dropped
`aixplain/v2/file.py:195`

`validate_file_exists` / `validate_file_size` no longer run, so size caps are not enforced before streaming. The backend presumably still rejects — this costs a wasted upload and a worse error message.

### N4. Empty recursive-children response clobbers `children`
`aixplain/v2/file.py:324`

A 200-with-empty-list recursive-children response overwrites `obj.children` with `[]`, discarding children the initial GET returned. Requires a specific backend response shape to trigger.

### N5. Temp ZIP leaked on mid-stream failure
`aixplain/v2/file.py:375`

The folder-download temp ZIP is created with `delete=False` outside its `try/finally`, so a mid-stream failure leaks it. Disk hygiene only.

### N6. Empty-list error body yields an empty message
`aixplain/v2/client.py:366`

The branch itself is a genuine fix — a list body previously hit `error_obj.get(...)` → `AttributeError`. The wrinkle is that an empty-list body now produces an empty message instead of falling back to `response.text`.

---

## Test coverage

All 1276 v2 unit tests pass, but every blocking finding is invisible to the suite because the mocks return minimal responses. **B1–B4 would all be caught by a single test** that echoes a realistic backend payload (including `url` and `s3Url`) back through `save()`, then calls `save()` a second time.

elsheikhams99 (approved):

PR #1057, opened 2026-08-18
feat(plugins): add aixplain-marketplace Claude Code plugin

nurhamdan1987:
Adds a `plugins/` tree as a sibling to `skills/`, plus the first plugin: **`aixplain-marketplace`**.

A plugin bundles a skill together with the MCP servers it needs, so one install gives a coding agent both the knowledge and the live connection:

```
/plugin marketplace add aixplain/aiXplain
```

## What it does

Wires up the first-party Marketplace Search MCP server (tool `6960f934f316da19e5f22494`, stable across environments) and ships a `marketplace-search` skill that drives it — find agents, models, tools, and integrations in one call, filter by category / developer / supplier / host / function, read pricing and hosting, then turn a hit into working code: SDK call, REST call, MCP config, or an `aix.Agent(tools=[...])` attach block.

The skill builds snippets from each asset's real input schema via `list_inputs_*` rather than emitting placeholders, and uses keyword arguments on `run()`. Verified against Cloud Translation on PROD:

```python
aix.Model.get("66aa869f6eb56342c26057e1").run(
    text="Good morning", sourcelanguage="en", targetlanguage="ar")
# SUCCESS
```

## Known-broken paths are documented, not retried

Two things do not work today, and the skill says so explicitly so an agent does not waste calls on them:

- **MCP `run_*` actions** — a single `input` string cannot express multi-field inputs. Execution is routed through generated SDK code instead.
- **Sorting** — `search` has no sort parameter, so "cheapest" requires paging the filtered set and sorting locally.

Both are filed separately with reproductions.

## Notes for review

- No secrets committed — `.mcp.json` references `${AIXPLAIN_API_KEY}`.
- Every documented call was verified against PROD before being written down; nothing is inferred from the SDK source.
- The skill is usable standalone (copy `SKILL.md` to `~/.claude/skills/`) for clients that read skills but not Claude Code plugins. The plugin README documents the Codex and Claude Desktop config too.
- New files use "aixplain" lowercase; the surrounding repo prose uses "aiXplain". Happy to align either way — flagging rather than silently changing existing files.
- `skills/README.md` says to open PRs into `main`; this one targets `development` as requested. Say if it should be retargeted.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,0 +1,22 @@
+{
+  "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
+  "name": "aixplain",
+  "description": "Claude Code plugins for the aixplain platform — search the marketplace and work with aixplain assets directly from your editor.",
+  "owner": {
+    "name": "aixplain",
+    "url": "https://aixplain.com"
+  },
+  "plugins": [
+    {
+      "name": "aixplain-marketplace",
+      "description": "Search the aixplain marketplace from your editor — find agents, models, tools, and integrations across the whole catalog, read pricing and hosting, and get working code to call any asset via SDK, REST, or MCP.",
+      "source": "./plugins/aixplain-marketplace",
+      "category": "development",
+      "author": {
+        "name": "aixplain",
+        "url": "https://aixplain.com"
+      },
+      "homepage": "https://studio.aixplain.com"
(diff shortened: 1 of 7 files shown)


nurhamdan1987:
Superseded by the combined Marketplace and Agent Builder PR: https://github.com/aixplain/aiXplain/pull/1058

PR #1061, opened 2026-08-25
docs(agent): clarify ContextOverflowStrategy (context condensation, not history replacement)

nurhamdan1987:
The `ContextOverflowStrategy` docstring (source of the generated API reference) inaccurately said `SUMMARIZE` "replaces the full chat history". Rewritten to be accurate and practical; the generated `agent.md` block is updated to match.

## What it now says
- Context condensation manages the **working context** sent to the model on a run — **not** Shared Memory or the stored session history, which is always retained.
- `TRUNCATE` — **default**; removes the oldest unprotected turns until the context fits.
- `SUMMARIZE` — summarizes older context into a shorter form; retains more meaning than truncation but adds latency and model cost.
- Available in SDK 0.2.46+ (use the current 0.2.47).
- Precedence, highest first: per-run override (`execution_params`) → saved agent setting (`agent.context_overflow_strategy`) → Agent Engine default (`truncate`).

Verified against 0.2.47: enum values `truncate`/`summarize`, field default `None`, and `execution_params` maps `context_overflow_strategy` → `contextOverflowStrategy` (per-run override).

Companion narrative guide + examples: aixplain/sdk-documentation#264.

> Note: `docs/api-reference/python/aixplain/v2/agent.md` is generated by pydoc-markdown. I hand-updated the affected block to match the docstring since the generator isn't run in this environment; a full regen will reproduce equivalent output.

### DIFF
--- a/aixplain/v2/agent.py
+++ b/aixplain/v2/agent.py
@@ -126,11 +126,24 @@ class OutputFormat(str, Enum):
 
 
 class ContextOverflowStrategy(str, Enum):
-    """Strategy applied when input messages exceed the model's context window.
+    """Strategy for condensing the working context when a run exceeds the model's context window.
+
+    Context condensation shapes only the working context sent to the model on a
+    given run. It does not modify Shared Memory or the stored session history —
+    the complete session history is always retained.
 
     Attributes:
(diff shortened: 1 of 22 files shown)


aix-ahmet:
cherry picked to development

PR #1065, opened 2026-08-28
fix(v2): persist updated agent skills on save

xainaz:
Synchronize the public Agent.skills collection before validation and payload construction so replacements, mutations, and explicit clearing round-trip correctly. Add regression coverage for skill IDs, objects, clearing, and recursive saves.\n\nBUG-1166

### DIFF
--- a/aixplain/v2/agent.py
+++ b/aixplain/v2/agent.py
@@ -820,6 +820,39 @@ class Agent(
             ]
             self.files = current_ids
 
+    @staticmethod
+    def _skill_reference_id(skill: Optional[Union[str, Dict[str, Any], "Skill"]]) -> Optional[str]:
+        """Return the backend ID represented by one Skill reference."""
+        if skill is None:
+            return None
+        if isinstance(skill, str):
+            return skill
+        if isinstance(skill, dict):
+            return skill.get("id") or skill.get("asset_id") or skill.get("assetId")
+        return skill.id
+
+    def _sync_skill_references(self) -> None:
+        """Capture direct mutations to ``agent.skills`` before validation or save."""
+        current = list(self.skills or [])
+        original = list(getattr(self, "_original_skills", []) or [])
+        # A None slot is a still-unsaved placeholder; only safe to refresh by position if lengths match.
+        same_length = len(current) == len(original)
+        effective_current = [
+            original[index] if skill is None and same_length else skill for index, skill in enumerate(current)
+        ]
+        current_ids = [self._skill_reference_id(skill) for skill in effective_current]
+        original_ids = [self._skill_reference_id(skill) for skill in original]
+        if current_ids != original_ids or any(not isinstance(skill, str) for skill in effective_current):
+            original_by_id = {
+                self._skill_reference_id(skill): skill
+                for skill in original
+                if self._skill_reference_id(skill) is not None
+            }
+            self._original_skills = [
+                original_by_id.get(skill, skill) if isinstance(skill, str) else skill for skill in effective_current
+            ]
+            self.skills = current_ids
+
         # TODO: Re-enable this validation after backend data consistency is fixed
         # if self.agents and (self.tasks or self.tools):
         #     raise ValueError(
@@ -1301,6 +1334,7 @@ class Agent(
     def _save_subcomponents(self) -> None:
         """Recursively save all unsaved child components."""
         self._sync_file_references()
+        self._sync_skill_references()
         failed_components = []
 
         # Save tools
@@ -1359,6 +1393,7 @@ class Agent(
     def _validate_run_dependencies(self) -> None:
         """Validate that all child components are saved before running."""
         self._sync_file_references()
+        self._sync_skill_references()
         unsaved_components = []
 
         # Check tools
@@ -1404,6 +1439,7 @@ class Agent(
     def _validate_dependencies(self) -> None:
         """Validate that all child components are saved."""
         self._sync_file_references()
+        self._sync_skill_references()
         unsaved_components = []
 
         # Check tools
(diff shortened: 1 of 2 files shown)


MaiaJP-AIXplain:
I checked `ede5e1f8` against the base `23a6a0b8`, and I ran the round trip on DEV.

## How to reproduce

Every snippet below runs against this client. No network call is needed except where noted.

```python
import os
from aixplain import Aixplain

aix = Aixplain(api_key="test-key", backend_url="https://example.test")
```

## The bug and the fix

The base ignores every change to `agent.skills` after construction. The save payload drops the key:

```python
agent = aix.Agent(name="a", instructions="hi")
agent.skills = [aix.Skill(id="saved-2", name="Two")]
agent.build_save_payload().get("skills")   # base: None. This PR: the skill.
```

The fix works. On DEV I created an agent with the `data-analyst` skill, fetched it again, renamed it, and saved:

```python
live = Aixplain(api_key=os.environ["TEAM_API_KEY"], backend_url=os.environ["BACKEND_URL"])
agent = live.Agent(name="pr1065-review", instructions="You analyze data.",
                   skills=["6a7baf3df6aae4cf3a83a299"])   # data-analyst on DEV
agent.save()
fresh = live.Agent.get(agent.id)
fresh.name = "pr1065-review-renamed"
fresh.save()
print(live.Agent.get(agent.id).skills)
```

The skill survived:

```
after create, fetched skills: ['6a7baf3df6aae4cf3a83a299']
after unrelated save, skills: ['6a7baf3df6aae4cf3a83a299']
VERDICT: PRESERVED
```

The v2 suite gives 1286 passed.

## 1. The payload now always sends the skills key

`build_save_payload` sets `payload["skills"]` on every call. An agent that never touched skills now sends `skills: []`:

```python
agent = aix.Agent(name="plain-agent", instructions="hi")
payload = agent.build_save_payload()
"skills" in payload   # base: False. This PR: True, value []
"files" in payload    # both: False
```

The `files` block directly below yours guards this case with `_files_ever_configured`, and its own test states the rule:

> `test_agent_never_configured_with_files_omits_the_key`: "An agent that never touched files must not start sending files=[]."

The DEV round trip above shows no data loss today, because the GET response carries the skills and the SDK hydrates them. The guard still exists in this file for a reason. Please mirror it with a `_skills_ever_configured` flag, so an unrelated save never asserts an empty skills list.

## 2. A None placeholder can corrupt `_original_skills`

`__init__` stores `None` for an unsaved Skill. `_sync_skill_references` recovers that slot by position, and only when the list length matches. A later append changes the length, so the slot stays `None` and reaches the payload builder:

```python
agent = aix.Agent(name="a", instructions="hi")
agent.skills = [aix.Skill(name="Unsaved One")]
agent._validate_dependencies()          # raises, as designed. It also syncs.
# agent.skills is now [None], _original_skills is [Skill(id=None)]
agent.skills.append(aix.Skill(id="saved-2", name="Two"))
agent.build_save_payload()
# ValueError: A skill must be a Skill instance, a dict, or a skill id string.
```

The message names three accepted types, and the caller passed one of them. The real cause is the lost placeholder.

`files` reproduces this exactly, with its own message. So the hole belongs to the copied pattern, not to your change alone. It is new for skills, because the base ignored the assignment. The `files` `__init__` comment already names the better approach: "Unsaved files keep the object itself, not None, so list edits never lose track of which is which." The `_sync_*` methods then write `None` back. Please keep the object in both places, or handle a length change instead of skipping recovery.

## 3. A comment block now trails the wrong method

The two new methods land between `self.files = current_ids` and the commented-out validation. That block reads:

```python
        # TODO: Re-enable this validation after backend data consistency is fixed
        # if self.agents and (self.tasks or self.tools):
```

It concerns agents, tasks, and tools. It now sits at the end of `_sync_skill_references`. Please move it back to `_sync_file_references`, or lift it out of both.

## 4. Nit: the change predicate differs from the sibling

`files` tests `any(isinstance(file, (File, dict)) ...)`. Skills tests `any(not isinstance(skill, str) ...)`, which is also true for a `None` placeholder. The sync then rewrites state on every call while a placeholder exists. The result is the same, so this is cosmetic. Matching the sibling keeps the two readable side by side.

## Aside, unrelated to this PR

`aix.Agent.delete("<id>")` fails on the class:

```
ResourceError: Failed to delete resource: 'str' object has no attribute '_ensure_valid_state'
```

`aix.Agent.get(id).delete()` works. This deserves its own issue.

MaiaJP-AIXplain (approved):
Approved. I verified `9db8eb85` against the base `23a6a0b8`, and I ran the round trip on DEV again.

All four points are addressed:

1. `build_save_payload` now guards on `_skills_ever_configured` and pops the key otherwise. An agent that never touched skills omits the key again.
2. `__init__` and `_sync_skill_references` now keep the unsaved Skill object in the public list. The `None` placeholder is gone, so the payload builder no longer reports a wrong type. The error now reads "All skills must be saved before saving the agent," which names the real cause.
3. The validation comment is back at the end of `_sync_file_references`.
4. The change predicate now matches the sibling: `any(isinstance(skill, (Skill, dict)) ...)`. `Skill` is a runtime import at line 19, so the check is safe.

Checks I ran on the head commit:

```
1 assign-after-construction: [{'id': 'added-1', ...}]   # BUG-1166 stays fixed
2 replacement:               ['replacement']
3 explicit clear:            []
4 empty no-op:               <absent>
5 append to existing:        ['orig', 'added-2']
```

Live on DEV, create then fetch then rename then save: `VERDICT: PRESERVED`. The outbound skill object now carries the real name and description instead of `name: None`.

`tests/unit/v2` gives 1288 passed.

The `aix.Agent.delete("<id>")` failure I noted stays open and is not part of this PR.

PR #1066, opened 2026-08-28
fix(trigger): hydrate flat schedule attributes

xainaz:
Populate friendly Trigger schedule fields from backend configuration during rehydration, including recurrence and day mappings. Preserve top-level nextRunAt precedence while supporting a nested fallback, and exclude the nested read-only value from save payloads. Add focused coverage for once, daily, weekly, monthly, and recurring response shapes.

### DIFF
--- a/aixplain/v2/trigger.py
+++ b/aixplain/v2/trigger.py
@@ -114,7 +114,7 @@ class TriggerConfiguration:
     days_of_month: Optional[List[int]] = field(default=None, metadata=dj_config(field_name="daysOfMonth"))
     run_at: Optional[str] = field(default=None, metadata=dj_config(field_name="runAt"))
     start_at: Optional[str] = field(default=None, metadata=dj_config(field_name="startAt"))
-    next_run_at: Optional[str] = field(default=None, metadata=dj_config(field_name="nextRunAt"))
+    next_run_at: Optional[str] = field(default=None, metadata=dj_config(field_name="nextRunAt", exclude=lambda x: True))
     repeat: Optional[TriggerRepeatRule] = None
 
 
@@ -399,8 +399,11 @@ class Trigger(
         if trigger_type == "external":
             payload["triggerId"] = self.trigger_id
         elif self.configuration is not None:
+            if self.id and (self.run_at is not None or self.every is not None or self.start_at is not None):
+                # Hydrated flat schedule fields (at/every/on/...) are the editable
+                # surface; rebuild so an edit made after fetch is actually sent.
+                self.configuration = self._build_configuration()
             configuration = self.configuration.to_dict()
-            configuration.pop("nextRunAt", None)
             payload["configuration"] = _strip_none(configuration)
         else:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


aixplain/v2/trigger.py
```diff
@@ -241,6 +243,34 @@ def __post_init__(self) -> None:
             self.trigger_type = "time"
             self.configuration = self._build_configuration()
 
+    def _hydrate_schedule_fields(self) -> None:
```
aix-ahmet, line 246: Hydration creates a second copy of the schedule that reads live but is dead on write. Since `__post_init__` returns early on the `id`-set path, `configuration` is never rebuilt from the flat fields, so:

```python
t = aix.Trigger.get(trigger_id)   # daily at 09:00
t.at = "10:00"
t.save()                          # succeeds — but still fires at 09:00
```

`build_save_payload` re-sends the old `configuration` (`time: "09:00"`) and the edit is silently lost (verified locally; same for `every`, `on`, `run_at`, `interval`, `timezone`). This gap technically predates the PR, but before it these attributes read `None`, so they didn't look like a live API — hydrating them makes assigning to them the natural way to update a schedule.

Suggestion: for time triggers with `id` set, rebuild via `self._build_configuration()` in `build_save_payload` when any flat schedule field is set (the hydration here inverts `_build_configuration` cleanly for all five types, so an unchanged trigger still round-trips identically) — or, at minimum, document the flat attributes as a read-only snapshot of `configuration`.

aixplain/v2/trigger.py
```diff
@@ -114,6 +114,7 @@ class TriggerConfiguration:
     days_of_month: Optional[List[int]] = field(default=None, metadata=dj_config(field_name="daysOfMonth"))
     run_at: Optional[str] = field(default=None, metadata=dj_config(field_name="runAt"))
     start_at: Optional[str] = field(default=None, metadata=dj_config(field_name="startAt"))
+    next_run_at: Optional[str] = field(default=None, metadata=dj_config(field_name="nextRunAt"))
```
aix-ahmet: Read-only exclusion is now split across two places: this field serializes normally, and `build_save_payload` pops `nextRunAt` as a special case. Any other serialization of the configuration still emits it — `t.to_dict()["configuration"]` contains `nextRunAt` (verified) — so a future call site that builds a payload from `to_dict()` (e.g. the `BaseResource.build_save_payload` default) would send it and trip the backend's `forbidNonWhitelisted` validation.

The Trigger-level read-only fields already solve this with metadata:

```python
next_run_at: Optional[str] = field(default=None, metadata=dj_config(field_name="nextRunAt", exclude=lambda x: True))
```

`exclude` only affects encoding — decode/hydration keeps working (the top-level `next_run_at` field already relies on this). That lets you drop the `configuration.pop("nextRunAt", None)` in `build_save_payload`.

aix-ahmet (commented):
The hydration logic itself checks out: I verified all five schedule types (once/daily/weekly/monthly/recurring) hydrate correctly, top-level `nextRunAt` takes precedence over the nested value with fallback working, an unchanged fetch -> save round-trip produces an equivalent configuration payload, and the raw response dict is never mutated. All 42 trigger unit tests pass at the PR head. Two findings, neither blocking:

1. **Hydrated flat fields are readable but write-dead** — editing `at`/`every`/`on`/etc. on a fetched trigger is silently discarded on `save()` (inline at `_hydrate_schedule_fields`).
2. **Nested `nextRunAt` exclusion is patched at one call site instead of at the field** — `to_dict()` still emits it; a field-level `exclude` matches the existing pattern and removes the special-case pop (inline at the new field).

aix-ahmet (approved):

PR #1070, opened 2026-09-01
Eng 3489 sdk static graph

xainaz:


aix-ahmet:
## Review (rebased onto `development`)

I rebased the branch onto current `development` (now including #1136 and #1142) and force-pushed it as `deac7d2c`. The only conflict was in `aixplain/v2/__init__.py`; I kept development's trigger/file imports and added the graph exports. The PR is mergeable again, pre-commit CI is green, and the full unit suite passes locally (2350 passed). Findings below were each reproduced against the rebased head.

### ⚠️ Rotate the API key first

The first commit on this branch added `notebooks/static_graph_end_to_end.ipynb` with a real platform key (`API_KEY = "b65abf3d…"`, line 74 of that file at `02a56a7b`). The second commit only deletes the file, so the key is still readable in this branch's history on GitHub. Please rotate it. The repo only allows squash merges, so it will not reach `development` history, but it is exposed now.

### Blocking

**1. The first `save()` wipes the graph** — [agent.py#L715](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/agent.py#L715)
`graph`/`strategy` are declared with `exclude` + `decoder`, which `BaseResource._create` treats as "re-hydrate from the POST response" ([resource.py#L247](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/resource.py#L247)). A create response that does not echo `graph` leaves `agent.graph is None` and `agent.strategy is None`, and the next `save()` sends no graph. Reproduced with a mocked `{"id": "AG1", "name": "x", "status": "draft"}` response. The doc's "fetch and update" flow (`agent.graph.max_loop_iterations = 5`) then raises on a just-saved agent. `tools` is explicitly preserved across the create response; `graph` needs the same.

**2. An SDK-serialized graph cannot be read back when a node has no name** — [graph.py#L207](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L207), [graph.py#L603](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L603)
`Node.to_dict()` drops a `None` name, but `from_dict()` passes the node dict straight into the dataclass, where `name` is a required positional. `Graph.from_dict(Graph(entry_point=n, nodes=[n]).to_dict())` with `LLMNode(name=None, node_id="b", model="m")` raises a bare `TypeError`.

**3. `Agent.get()` fails for any non-static-graph `strategy`** — [agent.py#L802-L811](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/agent.py#L802-L811)
A GET response carrying `strategy: {"type": "dynamic"}`, or a `graphVersion` without a graph, raises inside `__post_init__` and becomes `ResourceError: Failed to deserialize Agent`. If the backend starts returning the new `strategy` field for ordinary agents (the Jira scope adds it to the agent record), every non-graph fetch breaks. The read path should ignore or tolerate these, not raise.

**4. `Agent.get()` also fails when the stored graph fails local validation** — [graph.py#L590](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L590), [graph.py#L365](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L365)
`from_dict` calls `validate()`, so a stored graph with an unreachable node, or one containing a `parallel`/`subgraph` node (the engine has eight node types; `NODE_TYPES` models six), makes the agent unfetchable and uneditable. The backend is authoritative; validate on save, be lenient on read.

### Should fix

**5. Graph edits are invisible to `is_modified`.** The field is excluded from serializable state, so `agent.graph.max_loop_iterations = 5` leaves `is_modified == False`. `save()` does not gate on it today, so the payload is still correct, but the property lies.

**6. Post-construction assignment is not coerced** — [agent.py#L802](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/agent.py#L802). `agent.graph = graph.to_dict()` or `agent.strategy = {"type": "static_graph"}` fails at save time with `AttributeError`, while `budget`/`tasks` are coerced in `__setattr__`. Setting `agent.graph = None` leaves stale `strategy` and `graph_version` behind, violating the `__post_init__` invariant.

**7. `graphVersion` is hardcoded** — [agent.py#L2361](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/agent.py#L2361). `Agent(graph=g, graph_version="2").build_save_payload()["graphVersion"]` is `"1"`.

**8. Two budgets in one payload** — [graph.py#L421](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L421). Agent-level `budget` (`maxCost`, camelCase) and `strategy.budget` (`max_cost`, snake_case) both go out with no reconciliation or warning, so the caller cannot tell which cap applies.

**9. Plain-data rule (AGENTS.md, landed after this PR opened).** `Graph.from_dict`/`Edge.from_dict` silently ignore unknown keys; node and `RetryPolicy` dicts raise bare `TypeError` on unknown keys ([graph.py#L603](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L603)); non-numeric `timeout` raises `TypeError` ([graph.py#L186](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L186)); a dict `retry_policy` raises `AttributeError`; an edge without `target` raises `KeyError` ([graph.py#L407](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L407)). There are no `TypedDict`s and nothing is classified in `docs/v2-plain-data.md`. All of these should be `ValidationError` naming the accepted fields, with rows added to the doc.

**10. Docs cannot run as written** — [static-graph-agents.md#L26](https://github.com/aixplain/aiXplain/blob/deac7d2c/docs/static-graph-agents.md#L26). Every example imports the context-less module-level `Agent` and calls `agent.save()` without an `Aixplain(...)` client. Use `aix = Aixplain(); aix.Agent(...)` like the rest of the docs.

**11. Low: duplicate-condition check compares JSON strings** — [graph.py#L524](https://github.com/aixplain/aiXplain/blob/deac7d2c/aixplain/v2/graph.py#L524). `Condition.equals("k", 1)` and `Condition.equals("k", 1.0)` are equal in Python but pass the duplicate check, leaving a dead edge.

**12. Test gaps** — [test_graph.py#L271](https://github.com/aixplain/aiXplain/blob/deac7d2c/tests/unit/v2/test_graph.py#L271). Nothing exercises the fallback-ordering rule, multiple unconditional edges, unknown edge endpoints, `entry_point` not a node, `StaticGraphStrategy.from_dict` rejection, the create/get rehydration path, or `is_modified` — which is exactly where items 1–6 live.

### Checked and fine
Kahn's cycle detection and BFS reachability (multi-edges, self-loops, entry point inside a cycle), unconditional-after-conditional ordering, falsy values surviving `Node.to_dict()` (`share_session=False`, `timeout=0`), `clone()` deep-copying the graph, `Agent.to_dict()` on Python 3.9 / dataclasses_json 0.6.7, and `run()` needing nothing graph-specific.

PR #1074, opened 2026-09-07
fix(v2): SEV2 correctness fixes for search filters, results, save() and run kwargs (BUG-1092, BUG-1094, BUG-1093, BUG-1091)

aix-ahmet:
## Summary

This branch carries four SEV2 correctness tickets, merged in dependency order because all four touch `aixplain/v2/resource.py` and `aixplain/v2/model.py`. Every ticket was reviewed and CI-checked as its own pull request before being merged down into this one; those PRs are linked below and their full write-ups remain readable there.

## Tickets in this branch

|---|---|---|
| [BUG-1092](https://aixplain.atlassian.net/browse/BUG-1092) | #1074 (this PR) | Search filters sent the enum member repr (`"SortBy.NAME"`) instead of its value, so ownership and sort filters never applied; sort direction is now ±1 for both enum and string inputs, and supplier filtering is normalised. |
| [BUG-1094](https://aixplain.atlassian.net/browse/BUG-1094) | #1077 | Partial results and failures were reported as complete successes: unparseable records no longer vanish while `total` still counts them, a missing envelope key uses the computed default rather than raising, an errored or truncated stream reports a non-SUCCESS status, and a polled empty or zero result is no longer coerced to `{}`. Item 4 (the v1 streamer) was skipped because v1 is unmaintained. |
| [BUG-1093](https://aixplain.atlassian.net/browse/BUG-1093) | #1079 | `save()` on a deleted resource raises instead of creating a duplicate, a create whose response fails to hydrate still records the server id so a retry cannot orphan it, and `save()` no longer writes SDK defaults over server-side fields the hydrating response omitted. |
| [BUG-1091](https://aixplain.atlassian.net/browse/BUG-1091) | #1084 | Caller-supplied run values are honoured or stripped rather than silently discarded: `stream=True` dispatches, `api_key` never rides in the model input body, a per-run override cannot delete a session's persisted spend cap, `run_stream` sends the same identity headers as `run`, and progress kwargs stay off the wire. |

- [BUG-1092](https://aixplain.atlassian.net/browse/BUG-1092) — Search filters serialize Python repr() strings, so ownership/sort filters never apply
- [BUG-1094](https://aixplain.atlassian.net/browse/BUG-1094) — Partial results and failures are reported as complete successes
- [BUG-1093](https://aixplain.atlassian.net/browse/BUG-1093) — save() can resurrect a deleted resource or orphan a duplicate, and silently drops server-side fields
- [BUG-1091](https://aixplain.atlassian.net/browse/BUG-1091) — Caller-supplied values are accepted and silently discarded

## Notes for review

**Deliberate deviation (BUG-1091, acceptance criterion 4):** unknown run kwargs are still forwarded rather than rejected. Model parameters are declared per-model by the backend and ship without an SDK release, so a closed allowlist would reject valid parameters. The reasoning and the follow-up shape are recorded in #1084.

**Open question (BUG-1093):** whether the backend PUT is merge- or replace-semantics determines the blast radius of `save()` omitting unmodelled fields. Worth one backend check before sizing it; details in #1079.

## Testing

## Original ticket write-up

---
## Summary

`OwnershipType`, `SortBy`, `SortOrder` and `Supplier` are all declared as `class X(str, Enum)`. For those, `str(member)` returns the member *repr* (`"OwnershipType.PRIVATE"`), not the wire value (`"PRIVATE"`). The v2 search filter builders called `str()` eagerly, so passing the SDK's own enums to the SDK's own `search()` produced filters the backend could not match — the request came back **HTTP 200 with the unfiltered catalogue**, which makes the bug easy to miss.

## Changes

**`aixplain/v2/resource.py`** — three module-private helpers, now the single place this normalization lives:

- `_filter_value` — string enums, dict-valued enums (legacy v1 `Supplier`, via `code`/`id`), and plain strings.
- `_filter_values` — same, for filters the backend expects as an array; a scalar is wrapped in a one-element list, so `Supplier.OPENAI` and `[Supplier.OPENAI, "cohere"]` both work.
- `_sort_direction` — returns the contractual ±1 integer. Handles v2's `"ASC"`/`"DESC"`, the `"ASCENDING"`/`"DESCENDING"` spellings, and already-numeric input (v1's `SortOrder` is valued `1`/`-1`) mapped by sign, so a caller migrating from v1 keeps the direction they asked for.

`_populate_base_filters` uses them for `ownership`, `sortBy` and `sortOrder`. A sequence of ownerships now serializes to a list of values instead of one stringified list; `list`, `tuple` and `set` are all accepted.

**`aixplain/v2/model.py`**

- `functions`, `vendors`, `status`, `source_languages` and `target_languages` route through the shared helper, replacing ten near-identical copies of the `hasattr(x, "value")` idiom that was the root cause. The `vendors` guard in particular was written against v1's dict-valued `Supplier` and fell through to `str(s)` for v2's string-valued one, sending `"Supplier.OPENAI"`.
- `sort[].dir` is now ±1 for both enum and string inputs; it was previously the string `"DESC"` for an enum and `-1` for the equivalent plain string.
- An explicit `sort_by=None` / `sort_order=None` no longer serializes as the literal string `"None"` — each half independently falls back to its documented default (`"name"` / ascending).
- Deleted `find_supplier_by_id`: it called `.value.get("id")` on a `str`, so it raised `AttributeError` on every call. It had zero consumers. Regenerated docs entries removed alongside it.

**`aixplain/v2/skill.py`** — `suppliers` normalized the same way.

## Testing

The only existing test on this path passed plain strings, so the enum path had never been exercised. New `tests/unit/v2/test_search_filters.py` (308 lines) asserts the exact `client.request` body for enum inputs across five resource types, plus the numeric/spelling variants of sort order and the `None` defaults.

- `python -m pytest tests/unit` → **2026 passed, 3 skipped**
- `ruff check` + `ruff format` clean; pre-commit (including its pytest hook) passed on commit.

## Notes

The helpers are module-private (`_`-prefixed) and additive — no public API changes beyond the removal of the always-broken `find_supplier_by_id`.

## Jira

https://aixplain.atlassian.net/browse/BUG-1092

[BUG-1092]: https://aixplain.atlassian.net/browse/BUG-1092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BUG-1094]: https://aixplain.atlassian.net/browse/BUG-1094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BUG-1093]: https://aixplain.atlassian.net/browse/BUG-1093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -350,6 +350,30 @@ Beyond the factories, three more legacy prefixes redirect into v1:
 | `from aixplain.modules import Agent, Model, ...` | `aix.Agent`, `aix.Model`, … — v2 resources replace the v1 domain objects |
 | `from aixplain.decorators import ...`, `aixplain.base`, `aixplain.processes` | Internal helpers with no public v2 counterpart |
 
+## v2 behavior notes
+
+### `search()` is strict about records it cannot parse
+
+`search()` (and the `list()` wrappers built on it) raises `ResourceError` if the
+API returns a record the SDK cannot deserialize, matching the long-standing
+contract of `get()`. Previously such records were logged and dropped while
+`Page.total` still counted them, so `len(page.results) < page.total` was a
+normal state and bulk loops skipped rows with no error.
+
+If you prefer to skip the bad records, opt in per call:
+
+```python
+page = aix.Model.search(strict=False)
+
+page.total    # never counts the records that were skipped
+page.skipped  # how many were skipped on this page
+```
+
+`strict=False` is a client-side option and is never sent to the backend. A
+resource class whose records are known to be heterogeneous can set
+`PAGINATE_STRICT = False` to make lenient listing its default; a per-call
+`strict=` always wins.
+
 ## Getting help
 
 - Open an issue at <https://github.com/aixplain/aiXplain/issues> — especially if a gap above blocks you; that feedback shapes the removal timeline.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -374,6 +374,37 @@ resource class whose records are known to be heterogeneous can set
 `PAGINATE_STRICT = False` to make lenient listing its default; a per-call
 `strict=` always wins.
 
+### A streamed run only reports `SUCCESS` when it finished
+
+`ModelResponseStreamer.status` used to become `SUCCESS` as soon as the
+connection ended, so an error event mid-generation, or a stream cut before its
+`[DONE]` marker, handed you a truncated answer that claimed to be complete. It
+now stays `FAILED` unless the stream terminated cleanly (a `[DONE]` marker, a
+terminal `finish_reason`, or a terminal `status` envelope), and an error event
+yields one final chunk carrying the reason:
+
+```python
+with model.run_stream(text="Explain LLMs") as stream:
+    for chunk in stream:
+        print(chunk.data, end="", flush=True)
+
+if stream.status != aix.ResponseStatus.SUCCESS:
+    ...  # the text above is partial
+```
+
+Iteration itself does not raise, so existing `for chunk in stream` loops keep
+working — but check `stream.status` (or `chunk.error_message`) before treating
+the concatenated text as a full answer.
+
+### A polled falsy result keeps its value and type
+
+`poll()` coerced any falsy `data` to `{}`, so a classifier that legitimately
+answers `0`, or a model whose correct answer is `""`, came back as `{}` — and
+`result.data.strip()` raised `AttributeError`. Falsy-but-present payloads
+(`""`, `0`, `False`, `[]`) now pass through unchanged, matching what the same
+model returns on the synchronous path. A genuinely absent (or `null`) `data`
+still arrives as `{}`.
+
 ## Getting help
 
 - Open an issue at <https://github.com/aixplain/aiXplain/issues> — especially if a gap above blocks you; that feedback shapes the removal timeline.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -352,27 +352,43 @@ Beyond the factories, three more legacy prefixes redirect into v1:
 
 ## v2 behavior notes
 
-### `search()` is strict about records it cannot parse
+### `search()` never reports more records than it returns
 
-`search()` (and the `list()` wrappers built on it) raises `ResourceError` if the
-API returns a record the SDK cannot deserialize, matching the long-standing
-contract of `get()`. Previously such records were logged and dropped while
-`Page.total` still counted them, so `len(page.results) < page.total` was a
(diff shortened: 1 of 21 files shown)


aix-ahmet:

Reviewed at 21 files, +2773/-218. Full v2 suite passes (1480 passed, 4 skipped). Premises were checked rather than taken from the comments: `str(member)` does yield `"OwnershipType.PRIVATE"` for the `(str, Enum)` types, `api_key` really is inert because the client overlays `_auth_headers()`, `_server_fields`/`_explicit_roles` are invisible to `to_dict()`, and `build_run_url` never consumed `resource_path`.

|---|---|---|---|
| 1 | `aixplain/v2/agent.py:1307` | Medium | The deleted-save guard raises `ValidationError`, but the same guard in `BaseResource.save` (`resource.py:556`) sits inside `@with_hooks` and surfaces as `ResourceError`. They are sibling subclasses, so existing `except ResourceError` around `agent.save()` / `file.save()` (`file.py:298`) no longer catches it. The new tests assert the base class, which masks the split. |
| 2 | `aixplain/v2/resource.py:1178` | Medium | `PAGINATE_STRICT = True` makes one unparseable record abort an entire listing, so a single new unmodelled enum value breaks `Model.search()`, `Inspector.search()`, `APIKey.list()` and `Trigger.list()` outright. The SDK itself already needed `strict=False` at `agent_evaluator.py:1311`, which argues for a lenient default on discovery resources. |
| 3 | `aixplain/v2/resource.py:1143` | Medium | A missing items key becomes a silent empty page with only a warning, so a renamed envelope or a 200 error body is indistinguishable from "no results". That is inconsistent with the strict-by-default treatment applied to individual records in the same function. |
| 4 | `aixplain/v2/model.py:842` | Medium | `run(stream=True)` routes to `run_stream()` unless `supports_streaming is False`, but that field defaults to `None`. A model whose record omits `supportsStreaming` now gets an SSE parser over a plain JSON body, yielding raw JSON as `chunk.data` and `status=FAILED` where it previously returned a valid `ModelResult`. |

Minor, no action needed: `fallback_data is None` at `resource.py:1961` is dead code since data is already defaulted above, and `_build_resources` is no longer the hook `_build_page` uses, which is harmless today but silently moves the extension point.

PR #1076, opened 2026-09-07
fix: SEV2 HTTP, load, security and control-effectiveness fixes (BUG-942, BUG-1097, BUG-941, BUG-939, BUG-946)

aix-ahmet:
## Summary

This branch carries four SEV2 tickets from the SDK audit, merged in dependency order because each builds on the request/retry layer the previous one touched. Every ticket was reviewed and CI-checked as its own pull request before being merged down into this one; those PRs are linked below and their full write-ups remain readable there.

## Tickets in this branch

|---|---|---|
| [BUG-942 + BUG-1097](https://aixplain.atlassian.net/browse/BUG-942) | #1076 (this PR) | Jitter on every poll and backoff sleep, 429 added to the retry forcelist with Retry-After honoured, `stream_progress` given a deadline and an unconditional poll counter, connection pool sized explicitly, and the read timeout bounded by the remaining poll budget. BUG-1097's core item, POST removed from the urllib3 retry set, had already shipped in #1045; its residual items are here. |
| [BUG-941](https://aixplain.atlassian.net/browse/BUG-941) | #1080 | `call_run_endpoint` and `Pipeline.poll` returned a FAILED response object instead of raising `UnboundLocalError` on any network error or non-JSON body. |
| [BUG-939](https://aixplain.atlassian.net/browse/BUG-939) | #1083 | Ancestor `.env` files no longer hijack `BACKEND_URL` or the API key, debug logs no longer emit request bodies, response bodies or third-party tokens, every URL fetched from a caller or a response goes through a scheme and resolved-IP check, and the presigned upload host is validated before the PUT. |
| [BUG-946](https://aixplain.atlassian.net/browse/BUG-946) | #1085 | Five declared controls that did nothing now take effect: action-input validation reports its own crash, `progress_verbosity` reaches `sync_poll`, the API key is validated lazily so `--help` and unit collection work without credentials, the v2 client is constructed lazily instead of silently becoming `None`, and the team-agent invariant is resolved. |

- [BUG-942](https://aixplain.atlassian.net/browse/BUG-942) — Poll backoff has no jitter, stream_progress polls unbounded, no connection reuse, pool_maxsize=10
- [BUG-1097](https://aixplain.atlassian.net/browse/BUG-1097) — The BUG-938 timeout fix activated a latent POST-retry path (residual items; core fix shipped in #1045)
- [BUG-941](https://aixplain.atlassian.net/browse/BUG-941) — call_run_endpoint raises UnboundLocalError on any network error instead of the documented FAILED status
- [BUG-939](https://aixplain.atlassian.net/browse/BUG-939) — dotenv walk-up hijacks BACKEND_URL, debug logs dump bodies and tokens, three SSRF sinks, unvalidated upload URL
- [BUG-946](https://aixplain.atlassian.net/browse/BUG-946) — Five controls that do nothing: validation returns valid when it crashes, dropped parameter, CLI --help crashes, client silently None, invariant commented out

## Notes for review

**Scope note carried from BUG-947 review:** the functional-test CI leg still targets the production tenant. Retargeting it is a policy decision and was deliberately left out.

## Testing

## Original ticket write-up

---
## Summary

Every poll and backoff sleep in the SDK was deterministic. Any set of clients started together stayed phase-locked for the life of their runs and hit the platform in synchronized waves — and because this is a client SDK, the effect is multiplied by every installed copy, with the load landing on aiXplain's own platform.

This PR jitters every poll/backoff sleep, adds `429` to the retry forcelist, bounds `stream_progress` with a real deadline, and sizes the connection pools explicitly.

### v2 (primary)

**New `aixplain/v2/_backoff.py`** — one place for the retry/poll timing policy:
- symmetric multiplicative jitter (±20%)
- the SDK-standard `×1.1` / 60s-capped ladder
- a sleep clamped to the caller's remaining budget, so jitter can shorten a sleep but never overrun a deadline

**Jittered sleeps** — `sync_poll`, `_submit_with_retries`, `ActionMixin._poll_for_data` (which had a flat 1s tick and no backoff at all), and `stream_progress`.

**`client.py`**
- add `429` to `DEFAULT_RETRY_STATUS_FORCELIST`; set `backoff_jitter`, `backoff_max`, `respect_retry_after_header`, and a bounded `retry_after_max` (urllib3's own default is 6 hours). Retrying 429 cannot duplicate a billable submission because `allowed_methods` stays GET-only.
- optional kwargs are filtered against the *installed* `Retry` signature, so urllib3 1.26 loses the jitter rather than raising. Only the SDK's own optional hardening kwargs are dropped — a caller's typo still reaches `Retry` and raises rather than being silently discarded.
- size the `HTTPAdapter` pool explicitly (32/64, `block=False`); urllib3's 10/10 default discards connections above 10 in flight.

**`sync_poll` budget** (BUG-1097 second-order item) — each poll request is now bounded by its remaining budget, so one hung poll can no longer consume the whole timeout. `poll()` gains an additive `timeout` kwarg; a subclass still overriding the old signature loses the bound instead of breaking. The per-request read timeout is bounded against the *client's* configured timeout, so narrowing `AixplainClient(timeout=...)` is honoured rather than silently widened back to 300s. Budget exhaustion also wins over the transport error it causes — the last poll before the deadline failing now surfaces as the documented `TimeoutError`, not `APIError`. `UntrustedURLError` is never softened.

**`stream_progress`** — add a `timeout` with a wall-clock deadline, raise the default poll interval from 0.05s to 0.5s with backoff, and increment `_poll_count` unconditionally so `max_polls` can terminate a run that never emits steps.

**`upload_utils`** — reuse one pooled `RequestManager` session instead of building one per request.

### v1 (minimal, line-level only — v1 is unmaintained)

- Jitter the four v1 poll sleeps.
- `request_utils`: one module-level `Session` instead of one per call, with an explicitly sized pool (the 10/10 default would have cancelled out the reuse the shared session exists to provide).
- `pipeline`: default timeout 20000s → 1800s.
- Per-poll logs no longer interpolate the response body, and use lazy `%s` args so nothing is stringified when the level is off.

## Out of scope

BUG-942 item 6 (client-side rate limiting / token bucket).

## Testing

New unit tests covering jitter distribution and budget clamping, the 429/`backoff_jitter` retry configuration and its urllib3-1.26 fallback, `sync_poll` budget accounting and error precedence, `stream_progress` deadline/`max_polls` bounds, upload timeouts, and a guard test asserting no unjittered poll sleeps remain. `ruff check` and `ruff format --check` are clean on all changed files.

[BUG-941]: https://aixplain.atlassian.net/browse/BUG-941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BUG-939]: https://aixplain.atlassian.net/browse/BUG-939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,5 +2,26 @@ BACKEND_URL=https://platform-api.aixplain.com
 MODELS_RUN_URL=https://models.aixplain.com/api/v1/execute
 PIPELINE_API_KEY=<API_KEY>
 MODEL_API_KEY=<API_KEY>
-LOG_LEVEL=DEBUG
+# DEBUG is deliberately not the documented value: it used to log full request
+# and response bodies, including the aiXplain API key and any third-party token
+# inside a tool payload (BUG-939).
+LOG_LEVEL=INFO
 TEAM_API_KEY=<YOUR_API_KEY>
+
+# --- Optional security opt-ins (all default to off) ---
+# Log request/response bodies at DEBUG. Values are redacted and truncated, but
+# redaction is best-effort: do not enable it on a shared log sink.
+# AIXPLAIN_LOG_BODIES=1
+#
+# Allow plain http:// for BACKEND_URL / MODELS_RUN_URL / PIPELINES_RUN_URL and
+# for presigned upload URLs. Needed for local development against
+# http://localhost, and for an on-prem deployment terminating TLS elsewhere.
+# AIXPLAIN_ALLOW_INSECURE_URLS=1
+#
+# Allow the SDK to fetch code/artifact URLs that resolve to private or loopback
+# addresses. Only set this if your artifact host really is internal.
+# AIXPLAIN_ALLOW_PRIVATE_FETCH=1
+#
+# Extra hosts accepted as presigned upload targets, comma separated.
+# *.amazonaws.com and *.aixplain.com are always allowed.
+# AIXPLAIN_UPLOAD_HOST_ALLOWLIST=storage.example.com
--- a/.env.example
+++ b/.env.example
@@ -19,7 +19,8 @@ TEAM_API_KEY=<YOUR_API_KEY>
 # AIXPLAIN_ALLOW_INSECURE_URLS=1
 #
 # Allow the SDK to fetch code/artifact URLs that resolve to private or loopback
-# addresses. Only set this if your artifact host really is internal.
+# addresses. Only set this if your artifact host really is internal. The cloud
+# metadata endpoints stay blocked either way.
 # AIXPLAIN_ALLOW_PRIVATE_FETCH=1
 #
 # Extra hosts accepted as presigned upload targets, comma separated.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -369,6 +369,58 @@ Beyond the factories, three more legacy prefixes redirect into v1:
 | `from aixplain.modules import Agent, Model, ...` | `aix.Agent`, `aix.Model`, … — v2 resources replace the v1 domain objects |
 | `from aixplain.decorators import ...`, `aixplain.base`, `aixplain.processes` | Internal helpers with no public v2 counterpart |
 
+## Polling behaviour changes
+
+Two polling defaults changed. Both bound a loop that was previously unbounded or
+effectively unbounded; both are opt-out-able through the same parameter you
+already pass.
+
+### `Pipeline.run` / `Pipeline.poll` stop polling after 30 minutes
+
+The default `timeout` on `aixplain.modules.pipeline.Pipeline.run()` (and the
+private polling loop behind it) dropped from **20,000 seconds (5h 33m) to 1,800
+seconds (30 minutes)**.
+
+A pipeline that legitimately runs longer than 30 minutes will now stop being
+polled and be reported as a failure, even though the run itself continues on the
+platform. If you have such a pipeline, raise the budget explicitly:
+
+```python
+pipeline.run(data, timeout=20000.0)   # the previous default
+```
+
+The old default meant a pipeline that never completed pinned a thread for over
+five hours; 30 minutes is the bound for the common case, and the parameter is
+there for the rest.
+
+### `AgentProgressTracker.stream_progress` is bounded and raises on expiry
+
+`stream_progress` used to be a `while True` loop with no `timeout` parameter at
+all. It now takes `timeout` (default **300 seconds**) and raises
+`TimeoutError` when that budget expires with the run still non-terminal.
+
+This matches `sync_poll`, which has always raised `TimeoutError` in the same
+situation — the two polling surfaces previously disagreed about whether an
+expired budget was a result or an error.
+
+```python
+# previous behaviour: poll forever
+tracker.stream_progress(url, timeout=None)
+
+# a longer budget
+tracker.stream_progress(url, timeout=1800)
+
+# or handle the expiry
+try:
+    response = tracker.stream_progress(url)
+except TimeoutError:
+    ...
+```
+
+A terminal status (`SUCCESS`, `FAILED`, `ABORTED`, `CANCELLED`, `ERROR`) and the
+`max_polls` cap still *return* the response as before; only the wall-clock
+deadline raises.
+
 ## Getting help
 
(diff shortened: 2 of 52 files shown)


aix-ahmet:

Reviewed at 52 files / 27 source. Core of the PR (SSRF policy module, GET-only retries + 429 + jitter, body-redacted logging, v1 UnboundLocal fixes) verified sound. Open items below, most severe first.

|---|---|---|---|
| 1 | `aixplain/v2/file.py:310` | High | `File.save()` calls `validate_fetch_url(self.source)` then `requests.get(..., allow_redirects=True)`. A public host that 302s to `169.254.169.254` defeats the guard and uploads instance credentials. `aixplain/v2/file.py` is also absent from `GUARDED_SINK_FILES` in `tests/unit/v2/test_ssrf_call_sites.py`, so the test does not catch it. Use `safe_get`. |
| 2 | `aixplain/v1/modules/benchmark_job.py:117` | Medium | Same redirect bypass: `validate_fetch_url(csv_url)` then `save_file` → `_request_with_retry` follows redirects unvalidated. |
| 3 | `aixplain/v2/upload_utils.py:220`, `aixplain/utils/file_utils.py:250`, `aixplain/v2/file.py:218` | Medium | `validate_upload_url` does not survive a redirect: an allowed host answering `307` re-sends the file body to a host that never passed validation. Pass `allow_redirects=False` (presigned PUTs never legitimately redirect). |
| 4 | `aixplain/v2/resource.py:1628` | Medium | The `sync_poll` budget is still not enforced, so BUG-1097 is only partly closed. GET is in `RETRY_ALLOWED_METHODS`, so urllib3 retries a read timeout up to 5 more times below `requests`; `sync_poll(timeout=300)` against a hung endpoint can block ~30 min before re-checking its deadline. Use `remaining / (retry_total + 1)`, or poll through a session with retry total 0. |
| 5 | `aixplain/v2/agent_progress.py:897` | Medium | `stream_progress` now defaults to a 300s timeout and on expiry returns the last IN_PROGRESS response instead of raising. `sync_poll` raises `TimeoutError` in the same situation. At minimum this needs a MIGRATION.md entry. |
| 6 | `aixplain/v1/modules/pipeline/asset.py:99,220` | Medium | Undocumented 11x reduction of the pipeline polling budget (20000s to 1800s). A pipeline that legitimately runs over 30 minutes will start reporting failure. Unrelated to the jitter/logging changes in the same hunk. |
| 7 | `aixplain/utils/config.py:31` | Medium | `validate_config_url` runs at module import, so `BACKEND_URL=http://localhost:8000` makes `import aixplain` itself raise. This conflicts with the BUG-946 goal of keeping `--help` and unit collection working. Consider warning at import and failing at first request. |
| 8 | `aixplain/v2/tool.py:498` | Medium | `_validate_params` converts any exception into a blocking validation error, but the lazy input loader hits the network, so a transient backend 5xx now fails `tool.run()`. Distinguish unknown-action from transport failure. |
| 9 | `aixplain/utils/request_utils.py:21` | Low/Medium | The shared process-wide `requests.Session` is used from `ThreadPoolExecutor` pools in the v1 agent and team-agent factories. `Session` is not thread-safe, and every response mutates `session.cookies`. Consider a `threading.local()` session or blocking cookies. |
| 10 | `aixplain/utils/url_safety.py:377` | Low | `safe_get` drops headers only on a netloc change, so an https to http downgrade on the same host keeps the caller's headers in clear text. Compare scheme as well. |
| 11 | `aixplain/v2/resource.py:1776` | Low | A real 401/403 on the last poll is swallowed and reported as `TimeoutError`. Chain the original with `raise ... from e`. |

Verified correct, no action: the `__init__` module `__getattr__`; `_build_retry` optional-kwarg handling against urllib3 2.7; 429 in the forcelist being safe given GET-only retries; the `progress_verbosity` passthrough; lazy `%s` logging against `PipelineResponse`; and the `parametersTemp` / `resp` / `r.status_code` UnboundLocal fixes.

PR #1078, opened 2026-09-07
fix(tests, docs): SEV2 functional-test safety, run-metadata disclosure and v1 RAG URL contexts (BUG-947, BUG-1096, BUG-945)

aix-ahmet:
## Summary

This branch carries three SEV2 tickets covering test hygiene, documentation and one v1 retrieval bug. Every ticket was reviewed and CI-checked as its own pull request before being merged down into this one; those PRs are linked below and their full write-ups remain readable there.

## Tickets in this branch

|---|---|---|
| [BUG-947](https://aixplain.atlassian.net/browse/BUG-947) | #1078 (this PR) | Functional tests no longer mutate resources they do not own: the API-key test creates and cleans up its own key instead of randomising a pre-existing production key's rate limits, cleanup is scoped to resources created in the run rather than matched by a generic name, uploads are cleaned up, and teardown failures are reported instead of swallowed. |
| [BUG-1096](https://aixplain.atlassian.net/browse/BUG-1096) | #1081 | Run metadata is intentional, so this documents it rather than changing behaviour: what is collected (region, language, timezone, public IP and city-level coordinates from a single cached ipinfo.io lookup), why, and which call sites send it. To answer the question on the ticket, the direct v2 agent run path does send it; v2 runs routed through a session do not. |
| [BUG-945](https://aixplain.atlassian.net/browse/BUG-945) | #1082 | v1 RAG mode embedded the URL string instead of the document it points at, so retrieval silently never happened. One line, matching v2, plus tests. |

- [BUG-947](https://aixplain.atlassian.net/browse/BUG-947) — Functional tests mutate production: one rewrites the rate limits of a pre-existing prod API key
- [BUG-1096](https://aixplain.atlassian.net/browse/BUG-1096) — Every v1 agent run POSTs the caller's public IP, latitude and longitude to the backend
- [BUG-945](https://aixplain.atlassian.net/browse/BUG-945) — v1 RAG mode never fetches URL contexts: it embeds the URL string and answers from it

## Notes for review

**Left out deliberately (BUG-947):** the CI workflow still points the functional legs at the production tenant. Retargeting the tenant is a policy decision, not a test fix.

**Follow-up fix included:** moving `agent_validation_test.py` into the unit suite initially broke CI, because `AgentFactory.create` resolves the LLM before it validates the agent name. The lookup is now stubbed so the test exercises name validation with no network.

## Testing

## Original ticket write-up

---
## Problem

The functional suite runs against the **production tenant** on every push to `main`, and parts of it mutated or deleted live data it did not own:

- `apikey/test_api.py::test_05` took the first key `APIKeyFactory.list()` returned — whatever real key that happened to be — rewrote its global and asset limits to `randint(0, 10000)`, and never restored them.
- A module-scoped autouse fixture enumerated every key in the tenant and deleted any matching a generic name, with the failure swallowed.
- `agent_functional_test.py::test_agent_with_pipeline_tool` deleted *every* pipeline in the tenant matching the query `"Hello Pipeline"` before creating its own.
- Six copies of a `resource_tracker` fixture across four files ended teardown with `except Exception: pass` (two files defined it twice, the later silently shadowing the earlier), so a failed delete left an orphaned production asset with no trace anywhere. Leakage was invisible, not merely tolerated.

## Changes

**Scope every mutation to resources the test owns**
- `test_05` now creates the key it updates.
- API-key cleanup is scoped to keys created in this run: `TEST_API_KEY_NAME` carries a per-session uuid, and the delete goes through the object handle instead of a name sweep.
- The pipeline-tool test uses a unique name per run and registers the pipeline it owns.

**One tracker, with visible teardown** — `resource_tracker` moves into `tests/functional/conftest.py` as the single definition, backed by `tests/cleanup_guards.py`. The six local copies are deleted; the ~71 `.append()` call sites are unchanged because `ResourceTracker` subclasses `list`. Teardown now reports what it could not delete, fails the test by default (`AIXPLAIN_FUNCTIONAL_CLEANUP_STRICT=0` downgrades to warnings), and a session summary names every orphan. `mark_cleaned()` deregisters a resource the test deleted itself — most v1 deletes report one generic message for every failure, so a double delete is otherwise indistinguishable from an outage.

**Close the remaining leaks** — `file_create_test.py` no longer uploads with `is_temp=False` by default, since the SDK exposes no delete for a permanent upload; that path is opt-in behind `AIXPLAIN_ALLOW_PERMANENT_UPLOADS`. The `data_asset` onboarding tests register what they onboard *before* the polling loop.

**Make it enforceable** — five static guards in `tests/unit/test_functional_hygiene.py`: no list-then-mutate, no delete-by-generic-name, no unguarded permanent upload, one tracker definition, no newly swallowed cleanup. Each detector is tested against a synthetic tree, following the existing `tests/unit/test_ci_matrix_coverage.py`. `tests/unit/test_functional_cleanup.py` covers the tracker contract (deletion order, already-gone tolerance, failure aggregation, ledger/strict-raise wiring) in-process.

**Move two credential-free files out of the production-credentialled CI leg** — `general_assets/index_factory_test.py` and `agent/agent_validation_test.py` move to `tests/unit`. Every create in them sits inside `pytest.raises` and the SDK raises during argument validation, so they need neither a credential nor a network. The index-factory file is split so the one case that *did* reach the network (building `AirParams` runs a pydantic validator calling `ModelFactory.get`) has that lookup stubbed — otherwise it would have failed the credential-free `unit-coverage` job.

## Known limitation (not fixed)

`benchmark_functional_test.py` still leaks: it creates real benchmarks, and neither `Benchmark` nor `BenchmarkFactory` exposes a delete, so there is no cleanup to register. This is documented in place with a guard that fails the day a delete lands. (`sql_tool` and `test_rlm`, also on the ticket's list, create nothing on a backend.)

The leak summary is process-local under `pytest-xdist`.

## Verification

- `2048 passed, 3 skipped` (was `1972 passed, 3 skipped`)
- `tests/functional` collects `399` with no import errors (was `401`, less the two moved files)
- `tests/unit` passes with the environment stripped entirely, confirming the moved files are credential-free

## Jira

https://aixplain.atlassian.net/browse/BUG-947

[BUG-947]: https://aixplain.atlassian.net/browse/BUG-947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BUG-1096]: https://aixplain.atlassian.net/browse/BUG-1096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BUG-945]: https://aixplain.atlassian.net/browse/BUG-945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aix-ahmet:

Reviewed at 32 files. Verified in a worktree of the PR head: `tests/unit` gives 2097 passed with one pre-existing unrelated failure (a plotly import check, plotly not installed locally); the 80 new hygiene and cleanup tests, the new RLM tests, the moved validation tests and the run-metadata doc tests all pass; `pytest --collect-only tests/functional` collects cleanly.

The production-code changes check out. `_resolve_url_context` sits inside the `try` in v1 `_run_rag`, so a fetch failure degrades to FAILED exactly as `_run_parallel` does; the v1 and v2 call-site counts now match at two each; the only other v1 mode goes through the sandbox and correctly does not need it; and the metadata changes are documentation only, matching shipped behaviour.

All four findings are teardown-ordering issues that the old swallowing fixture hid and strict cleanup now exposes.

|---|---|---|---|
| 1 | `tests/functional/team_agent/team_agent_functional_test.py:218` | Medium/High | `new_agent` is registered after `team_agent` and is added to that team, while the later `agents.pop(0)` removes a different agent. Newest-first cleanup therefore deletes `new_agent` while the team still references it, and the resulting "currently used by one or more team agents" message matches none of the `_ALREADY_GONE` patterns, so teardown raises `CleanupError` on every run. Register `new_agent` before `team_agent`, or mark it cleaned and let the team delete cascade. |
| 2 | `tests/functional/apikey/test_api.py:71` | Medium | `api_key_leak_check` lists keys immediately after a function-scoped teardown deleted one, with no settle delay, though the code it replaces had an explicit half-second wait for exactly this reason. A stale entry gets deleted again, and `APIKey.delete()` collapses every failure including 404 into one message that `is_already_gone` cannot match, producing a false leak report on a clean run. Re-list or settle before deciding, or treat a leak-check delete failure as inconclusive. |
| 3 | `tests/functional/agent/agent_functional_test.py:160`, also `:814`, `:1022`, and `tests/functional/team_agent/team_agent_functional_test.py:570` | Low/Medium | The agent is registered before the tool it uses, so newest-first cleanup deletes the tool while the agent still lists it. The same file gets this right elsewhere, registering utility tools before the agent. Reorder these four sites to the dependency-last convention. |
| 4 | `tests/functional/agent/agent_functional_test.py:624` | Low | `resource_tracker.append(pipeline)` runs after both `save()` and `deploy()`. `save()` is what creates the pipeline, so a failing `deploy()` leaks it, which is the gap this PR closes elsewhere by registering before the risky step. Move the append to just after `save()`. |

Nothing else stood out. `cleanup_guards.py` handles the hard cases correctly: identity-based `mark_cleaned`, clearing before iterating so a retried cleanup cannot double-delete, preferring `delete()` over `__call__`, and wrapping `functools.partial`. Duplicate appends are tolerated because a second agent delete returns 404, and no dangling references to the removed helpers remain.

PR #1088, opened 2026-09-08
fix(v2): stop leaking the progress display thread on failed and concurrent runs (BUG-943)

aix-ahmet:
## Summary

The v2 agent progress display starts a daemon thread that repaints stdout at 20 Hz, and the only thing that ever stopped it was `AgentProgressTracker.finish()`. Four independent holes meant that call was routinely never made, so every failed, timed-out or concurrent agent run with `progress_format` set leaked a thread that kept printing `\r`-terminated lines for the life of the process — interleaving with and corrupting every other line on the same fd, including in non-TTY server containers where the output is pure log garbage.

All four holes were re-located and confirmed on this branch before any change, and none of the merged BUG-942 `stream_progress` work (deadline, `TimeoutError` on expiry, unconditional `_poll_count`) is undone.

## Design

**One frame owns the tracker.** `Agent.run`, `Agent.sync_poll` (standalone) and the session path each start the tracker, run the poll inside a single `try/except BaseException: tracker.stop(); raise`, and render the completion summary on success. The tracker reaches `on_poll` as the `_progress_tracker` run kwarg, a `_RUN_CONTROL_KEYS` entry that `_payload_kwargs_for_run` / `build_run_payload` strip before anything is posted. `run()` calls `sync_poll` with the key present, which tells `sync_poll` that an enclosing owner is driving the display and not to start a second one.

Consequences:

- `before_run` no longer starts anything and `after_run` no longer tears anything down, so an `after_run` override that skips `super()` cannot leak the thread.
- Nothing is shared between runs, on the instance or in a context, so a nested progress-less `run_async()` fired from a hook cannot adopt, stop or print the outer run's display. Two concurrent `run()` calls on one `Agent` get two independent trackers.
- There is no ContextVar, no reset token on the tracker, no `_unbind`/`_stop` helper pair and no `_progress_tracker` property. The dataclass field is deleted outright: `grep -rn _progress_tracker` finds no reader outside `agent.py`.
- `Agent.run` calls `super().run` directly, so the shared `build_run_payload` warning `stacklevel` still resolves to the caller's line on both `run()` and `run_async()`.

**`run_async` + `sync_poll`.** `run_async` returns as soon as the run is submitted; it cannot drive a display, so it starts none. It still accepts `progress_format` (nothing breaks) but logs a warning saying so and pointing at the replacement: pass the progress kwargs to the poll, `agent.sync_poll(r.url, progress_format="status")`, which owns its own tracker for the duration of the poll. This is a documented change from base, where `run_async(progress_format=...)` bound a tracker to the instance that the user's later `sync_poll` fed and nothing ever stopped. The `run_async` and `sync_poll` docstrings describe the new pairing.

**TTY gate on the animation only.** `isatty()` (or a notebook, or an explicit override) decides whether the 20 Hz repaint thread starts. The `logs` step timeline and the newline-terminated completion summary are ordinary caller-thread output and are written regardless, so `progress_format="logs"` under CI or `tee` still yields the full timeline and credits summary; `status` mode off a terminal yields just the summary line. Overrides: `AgentProgressTracker(force_display=…)` and `AIXPLAIN_PROGRESS_FORCE_DISPLAY`, the latter a pure opt-in — only `1`/`true`/`yes`/`on` count, `false` or `0` leave auto-detection in charge.

**Fresh stop event per `start()`.** A painter blocked in `print()` (full pipe, paused pager) can outlive the bounded join in `stop()`. `start()` used to `clear()` the one shared event, re-arming that survivor next to the new thread. Each `start()` now hands its thread its own `Event`; the survivor's stays set and it exits the moment `print()` returns.

**Render failures never discard a result.** A `UnicodeEncodeError` printing `✓` on an ascii console or a `BrokenPipeError` in the summary line is logged at WARNING and the completed, billed result is returned; the thread is still stopped.

**Widened hook contract.** `after_run` is reached on the failure path with the raised exception, which can be a `KeyboardInterrupt` or `SystemExit`. `RunnableResourceMixin.after_run`, its docstring and `docs/api-reference/python/aixplain/v2/resource.md` / `agent.md` / `llms-full.txt` now promise `Union[ResultT, BaseException]`, matching `Agent.after_run`, and say the return value is ignored on the failure path.

## Per-item status

| # | Suggested fix | Status | Where |
| - | ------------- | ------ | ----- |
| 1 | `_finish_progress_tracker` must always stop and join the thread | Fixed | `AgentProgressTracker.stop()` is the single unconditional exit. The owner calls it on any exception and `finish()` (which calls it first) on success. An errored run stops cleanly and prints no completion summary. |
| 2 | `RunnableResourceMixin.run` needs `try/finally` | Fixed | `run()` / `run_async()` wrap their body in `except BaseException → _apply_after_run_failure(exc) → raise`, so `after_run` sees every outcome. The hook's return value is ignored on the failure path and a raising hook cannot mask the original error. The progress display no longer depends on this: `Agent.run` owns it directly. |
| 3 | Make the tracker run-local | Fixed | Owned by the calling frame and passed as a run kwarg; no instance or context state at all. |
| 4 | Gate the display thread on `sys.stdout.isatty()` | Fixed | Thread start only. Caller-thread `logs` timeline and completion summary still print off a terminal. |
| 5 | Don't hold `_display_lock` across the sleep; use `Event.wait` | Fixed | The loop copies the snapshot under the lock, renders and waits outside it, parked on its own per-start `Event`. `stop()` is observed in microseconds. |
| — | `run_async` starts a tracker nothing ever stops | Fixed | It starts none; see the `run_async` + `sync_poll` section. |
| — | A second `start()` orphaned the first thread | Fixed | `start()` calls `stop()` first and binds a fresh `Event`, so neither a stopped nor a stuck predecessor is ever re-armed. |

## Behaviour changes (intentional)

- **`run_async(progress_format=...)` no longer produces a display.** It logs a warning; pass the progress kwargs to `sync_poll(result.url, ...)` instead. At base this pairing rendered a spinner and then leaked the thread.
- **Off a terminal, the spinner animation is suppressed; the `logs` timeline and the completion summary are not.** Opt the animation back in with `AIXPLAIN_PROGRESS_FORCE_DISPLAY=1` or `AgentProgressTracker(..., force_display=True)`.
- A run that raises no longer prints a completion summary. It never printed a correct one — `finish()` was skipped entirely; now the thread also stops.
- `RunnableResourceMixin.after_run` may receive any `BaseException`, not only `Exception`; hooks written to the old contract that test `isinstance(result, Exception)` should test `BaseException`.
- No public signature is broken. `force_display` is a new trailing optional parameter on `__init__`/`start`; `stop()` is new; `_progress_tracker` is a new private run kwarg; the `Agent._progress_tracker` field (private, `init=False`, excluded from serialization, no reader outside `agent.py`) is removed.

## Test evidence

`tests/unit/v2/test_progress_thread_lifecycle.py`, **59 tests**. Every thread assertion is bounded — joins with a timeout and fails naming the survivors, never sleep-and-hope. The TTY fixture patches the gate predicate rather than `sys.stdout`, because pytest re-assigns `sys.stdout` when it resumes global capture for the call phase.

- Raising run, timing-out run, submission failure and `KeyboardInterrupt` leave no display thread and restore `threading.active_count()`, on both the direct and the session path.
- `run_async(progress_format=...)` builds no tracker and logs the warning; `run_async` then `sync_poll(url, progress_format=...)` renders, is fed on every poll and stops; a failing standalone `sync_poll` stops; `sync_poll` inside `run()` starts no second tracker.
- An `after_run` override that skips `super()` cannot leak, on success or failure. A nested progress-less `run_async()` from inside the outer run's poll leaves the outer display alive, fed and stopped on time.
- Two concurrent `run()` calls on one `Agent` build two distinct trackers and leave neither thread alive.
- Off a terminal: `run(progress_format="logs")` prints the step timeline, the `✓ Completed` line and the credits, and starts no thread; `status` mode prints only the summary; `stream_progress` prints the timeline and summary with no thread. `force_display=True` and every truthy env value start the thread off a terminal; `false`, `0`, `no`, `off`, empty and junk neither disable a real TTY nor enable a pipe.
- A painter that survives a timed-out join is not re-armed by the next `start()`; the second `start()` still gets a live thread and both are dead after `stop()`.
- A `finish()` that raises `UnicodeEncodeError` before stopping its thread: the run returns the result, logs the warning and the thread is still stopped.
- `run()` cannot swallow the exception, substitute a result on failure, or let a raising `after_run` mask the original error; `after_run` may still substitute on success.
- Structural: `Agent` has no `_progress_tracker` attribute; `_progress_tracker` is in `_RUN_CONTROL_KEYS` and never appears in `build_run_payload` output.

```
$ python -m pytest tests/unit -q
2743 passed, 152 warnings in 35.16s

$ ruff check aixplain/v2 && ruff format --check aixplain/v2
All checks passed!
30 files already formatted

$ coverage report --include="aixplain/v2/agent_progress.py"
before (development): 565 stmts, 344 miss, 39%
after:                592 stmts, 221 miss, 63%
```

The remaining uncovered lines are verbosity-2/3 text formatting (`_get_label`, `_truncate_text`, `_print_step_details`, `_print_token_usage`, the `_format_step_line` branches), display formatting this PR does not touch.

## Follow-ups, not in this PR

- Should the display thread exist at all on the TTY path? Notebooks already repaint synchronously from `update()`, and the poll cadence is now ≥0.5 s after BUG-942. Deleting the thread would remove this bug class permanently at the cost of spinner smoothness.
- A suite-wide autouse leak guard asserting `threading.active_count()` is restored after every `tests/unit` test would catch the next such regression anywhere in the SDK, but risks flaking on unrelated tests that spawn threads.

Jira: https://aixplain.atlassian.net/browse/BUG-943

### DIFF
--- a/aixplain/v2/agent.py
+++ b/aixplain/v2/agent.py
@@ -1,6 +1,5 @@
 """Agent module for aiXplain v2 SDK."""
 
-import contextvars
 import json
 import logging
 import re
@@ -41,17 +40,6 @@ if TYPE_CHECKING:
 
 
 logger = logging.getLogger(__name__)
-
-
-# One progress tracker per *run*, never per Agent. It used to live on
-# ``self._progress_tracker``, so two concurrent ``run()`` calls on one shared
-# Agent overwrote each other: A's tracker was replaced by B's, and A's
-# ``finish()`` then stopped B's display thread while A's spun at 20 Hz until
-# process exit (BUG-943). A ContextVar is per-thread *and* per-asyncio-task, so
-# the slot is run-local by construction rather than by discipline.
-_ACTIVE_PROGRESS_TRACKER: contextvars.ContextVar[Optional[Any]] = contextvars.ContextVar(
-    "aixplain_active_progress_tracker", default=None
-)
 
 
 # Type definitions for conversation history
@@ -289,6 +277,9 @@ class AgentRunParams(BaseRunParams):
                         If None (default), progress tracking is disabled.
         progress_verbosity: Detail level - 1 (minimal), 2 (thoughts), 3 (full I/O)
         progress_truncate: Whether to truncate long text in progress display
+        _progress_tracker: Internal. The tracker owned by the ``run()`` /
+            ``sync_poll()`` call in progress, handed down to ``on_poll``.
+            Never sent to the backend.
     """
 
     session: NotRequired[Optional[Union["Session", Text]]]
@@ -308,6 +299,7 @@ class AgentRunParams(BaseRunParams):
     progress_format: NotRequired[Optional[Text]]
     progress_verbosity: NotRequired[Optional[int]]
     progress_truncate: NotRequired[Optional[bool]]
+    _progress_tracker: NotRequired[Optional[Any]]
 
 
 @dataclass_json
@@ -719,18 +711,6 @@ class Agent(
         metadata=config(field_name="contextOverflowStrategy"),
     )
 
-    @property
-    def _progress_tracker(self) -> Optional[Any]:
-        """Deprecated read-only alias for the active run's progress tracker.
-
-        The tracker moved off the instance and into a run-local ContextVar
-        (BUG-943); a per-instance slot is precisely what let concurrent runs
-        stop each other's display threads. Kept as a property so any
-        out-of-tree reader still sees a sensible value. Assignment is no longer
-        supported.
-        """
-        return _ACTIVE_PROGRESS_TRACKER.get()
-
     def __post_init__(self) -> None:
         """Initialize agent after dataclass creation."""
         self.tasks = [Task.from_dict(task) for task in self.tasks]
@@ -980,23 +960,22 @@ class Agent(
             return {"id": tool.get("id"), "type": tool.get("type")}
         return {"id": getattr(tool, "id", None), "type": getattr(tool, "type", None)}
 
-    def _start_progress_tracker(self, kwargs: Dict[str, Any]) -> Optional[Any]:
-        """Create, start and bind a run-local progress tracker.
+    # Run kwarg that carries the run's progress tracker from ``run()`` /
+    # ``sync_poll()`` down to ``on_poll``. Listed in ``_RUN_CONTROL_KEYS`` so
+    # it is stripped before the payload is built and never reaches the wire.
+    _PROGRESS_TRACKER_KWARG: ClassVar[str] = "_progress_tracker"
 
-        Returns the tracker so callers with real control flow (the session
-        path) can hand it straight back to :meth:`_finish_progress_tracker`;
-        the hook path resolves it from the ContextVar instead. Returns ``None``
-        when progress display was not requested.
-        """
+    def _start_progress_tracker(self, kwargs: Dict[str, Any]) -> Optional[Any]:
+        """Build and start a tracker from the progress kwargs; ``None`` if not requested."""
         progress_format = kwargs.get("progress_format")
         if progress_format is None:
             return None
 
         from .agent_progress import AgentProgressTracker, ProgressFormat
 
+        fmt = ProgressFormat(progress_format)
         progress_verbosity = kwargs.get("progress_verbosity", 1)
         progress_truncate = kwargs.get("progress_truncate", True)
-        fmt = ProgressFormat(progress_format)
 
         # ``poll_interval`` is deliberately not set: this tracker is driven by the
         # start/update/finish hooks off ``sync_poll``, which owns the interval.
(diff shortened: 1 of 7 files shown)


aix-ahmet:

The core leak is closed, but the fix introduces regressions in behaviour the PR does not disclose, and the ContextVar-based ownership model is the root of most of them. Each finding below was reproduced against the PR head, not inferred.

|---|---|---|---|
| 1 | `aixplain/v2/agent.py:1273` | High | `run_async` now stops the tracker in a `finally` as soon as the POST returns, so the documented `run_async(..., progress_format="status")` then `agent.sync_poll(r.url)` pattern renders nothing. At base the instance-bound tracker was driven by the user's `sync_poll` and showed a live spinner. Reproduced: PR head 0 stdout bytes, base 594 bytes. The in-code justification ("nothing would ever update the display") is factually wrong. |
| 2 | `aixplain/v2/agent.py:1239` | High | On the sync `run()` path, teardown depends entirely on the overridable `after_run` hook. An override that does not call `super()` (a pattern the base docstring invites and the PR's own tests use) leaks the thread and leaves a stale tracker in the thread's context that the *next* run silently adopts and drives. `run_async` got a defensive `finally`; `run` did not. |
| 3 | `aixplain/v2/agent.py:1059` | High | When `progress_format` is `None`, nothing is bound, so every `tracker=None` teardown falls back to `_ACTIVE_PROGRESS_TRACKER.get()` and adopts, stops and unbinds whatever *other* run's tracker is in the current context. A nested progress-less `agent_b.run_async()` from a hook kills `agent_a`'s display; the session variant prints A's completion line with B's result. At base, per-instance storage made cross-agent interference impossible. |
| 4 | `aixplain/v2/agent_progress.py:976` | High | The TTY gate suppresses *every* progress print off a terminal, including the caller-thread `logs` step timeline and the newline-terminated completion summary, not only the 20 Hz repaint thread that produced the log garbage. `progress_format="logs"` under CI or `tee` goes from a full timeline plus credits summary to zero bytes with no warning. Gate the thread start on `isatty`, not the `logs`/summary writers. |
| 5 | `aixplain/v2/agent_progress.py:828` | Medium | If `stop()`'s 1s join times out on a thread blocked in `print()` (full pipe, paused pager), the reference is dropped before the join, and the next `start()` clears the *shared* `_stop_display` Event, re-arming the stuck thread. Two painters then interleave on one fd and the survivor can never be joined. Bind a fresh Event per `start()`. |
| 6 | `aixplain/v2/agent_progress.py:208` | Medium | `AIXPLAIN_PROGRESS_FORCE_DISPLAY` treats any non-empty value as authoritative, so `=false` or `=0` silently disables rendering even on a real TTY or in a notebook, contradicting its own comment describing it as an opt-in. Only the `1` path is tested. |
| 7 | `aixplain/v2/resource.py:1981` | Medium | `after_run` now receives `BaseException` (KeyboardInterrupt, SystemExit) on a path that previously never reached it, but the base signature, docstring and generated docs still promise `Union[ResultT, Exception]`. A hook written to the documented contract misclassifies Ctrl-C as a result, raises, is swallowed into a WARNING, and skips its own teardown. Only `Agent.after_run` was widened. |
| 8 | `aixplain/v2/agent.py:1052` | Medium | Hook-error swallowing was added on the failure path but not the success path, so a render error in `tracker.finish()` (a `UnicodeEncodeError` printing `✓` on an ascii console, `BrokenPipeError` under force display) escapes `after_run` and turns a completed, billed run into an exception that discards the result. |
| 9 | `aixplain/v2/agent.py:1082` | Design | Tracker lifetime is owned by no single frame: started in `before_run`, torn down at four sites, glued together by a module ContextVar, a reset token stashed on the tracker, and a cross-context `ValueError` fallback. Having `Agent.run` own it in one `try/except` (the shape `_run_with_session` already has) and passing it to `on_poll` via a `_RUN_CONTROL_KEYS` kwarg removes all of that and is the root fix for findings 2 and 3. Viability verified against the current `_payload_kwargs_for_run` stripping and `sync_poll` kwarg forwarding. |
| 10 | `aixplain/v2/agent.py:722` | Low | The "deprecated read-only alias" `_progress_tracker` property is dead in-tree and answers a different question than the field it replaced: it returns the calling context's ContextVar value, not this instance's tracker. Delete it and its test. |

PR #1089, opened 2026-09-08
fix(cache): stop persisting the API key and harden the asset cache (BUG-940)

elsheikhams99:
The model cache wrote the account credential in cleartext, at mode 0644, into the current working directory -- confirmed at runtime, the file contained "api_key": "<TEAM_API_KEY>" with perms 0o644. Four compounding defects in one file, all reproduced before fixing.

1. Plaintext credential at rest. Model holds api_key as an instance attribute and add() serialized asset.__dict__, so the key landed on disk. Model now declares a __cache_fields__ allowlist mirroring what from_dict reads, and serialize() drops credential-shaped keys at any nesting depth as a backstop. Rebuilt models take the credential from the live session, so a rotated key is honoured and one caller's explicit api_key is never served to another.

2. use_cache=False still wrote. ModelFactory.get called cache.add() unconditionally on the opt-out path. Caching is now neither read nor written when it is declined, and constructing an AssetCache no longer touches the filesystem.

3. invalidate() unlinked the lock file while holding that lock, so two processes flocked different inodes and both entered the critical section -- measured 9 of 30 entries overlapping, ending in lost writes. Only the data file is removed now, and load() no longer invalidates from inside the lock. Verified 0 overlaps.

4. O(n) per call, O(n^2) to populate. Every get() rebuilt the cache and re-serialized it. The dominant cost was serialize() recursing into Enum.__objclass__, which expanded `supplier` into ~20 KB per model (one entry was 21,169 bytes; 500 models was a 10 MB file, not the ~1 MB estimated). Enums now serialize to their value, the instance is memoized process-wide, entries carry a precomputed payload, reads do no I/O, and there is an LRU cap.

Writes go through a private temporary file (0600) in a 0700 directory and are moved into place with os.replace, so there is no world-readable or truncated window and a serialization failure leaves the previous cache intact. Saves merge with the file under the lock instead of overwriting, so a concurrent add is not lost; add_list still replaces, as documented. Expiry is checked before deserializing, and also on read now that the instance outlives a single call.

The cache moved out of $CWD to $XDG_CACHE_HOME/aixplain (LOCALAPPDATA on Windows), matching default_experiment_cache_dir, overridable with AIXPLAIN_CACHE_FOLDER.

Agent loading was also an N+1: each build_tool issued its own ModelFactory.get, and the ThreadPoolExecutor bought nothing because the workers queued behind a global file lock. prefetch_tool_models resolves the whole batch, plus the main LLM, through the existing get_model_from_ids. A 20-tool agent goes from 22 requests to 1.

Also fixed along the way, in the same paths:

- add() stored asset.__dict__ rather than the asset, so a memoized cache would have returned a raw dict from ModelFactory.get.
- get_model_from_ids always fetched one extra page: the exit condition was checked after the request rather than before.
- get() returns a deep copy. The instance is now shared, and add_additional_info_for_benchmark mutates additional_info in place, which would have corrupted cached state for later readers.
- cache_utils.py had the same location and permission defects with no callers; it now shares the hardened helpers.

Measured: per-get wall time flat at 0.055 ms across 10/100/500 entries (was 1.5 / 4.0 / 8.8 ms); entries 21,169 -> 1,306 bytes.

Not fixed here, and worth its own ticket: load() rebuilds through Model.from_dict, so an LLM read back from disk is downgraded to a plain Model and loses its parameters. Doing it properly needs per-subclass serialization across every Model subclass. These changes shrink its reach -- in-process hits and prefetched models keep their real type.

Tests: 45 new unit tests in tests/unit/utils/asset_cache_test.py, one per acceptance criterion, including cross-process mutual exclusion. The existing functional test asserted the defect -- it required use_cache =False to create $CWD/.cache/model.json -- and is rewritten, with cases added for credential absence, permissions and the no-write path.

### DIFF
--- a/aixplain/utils/asset_cache.py
+++ b/aixplain/utils/asset_cache.py
@@ -3,20 +3,37 @@
 This module provides a generic caching system for aiXplain assets (Models, Pipelines,
 Agents, etc.) with file-based persistence, automatic serialization, expiration,
 and thread-safe operations.
+
+Security notes (BUG-940):
+    * The cache lives under the per-user cache directory (``$XDG_CACHE_HOME`` /
+      ``~/.cache/aixplain``), never the current working directory. A cache in
+      ``$CWD`` leaks into shared CI workspaces and Docker image layers.
+    * Cached entries are built from an explicit field allowlist and are additionally
+      scrubbed of credential-shaped keys, so the account API key is never written
+      to disk. :meth:`AssetCache.get` returns objects whose credential is taken from
+      the live configuration instead.
+    * Files are written atomically through a private temporary file
+      (``0600``) inside a ``0700`` directory, so there is no window in which a
+      world-readable or partially written cache file exists.
 """
 
-import os
-import logging
+import copy
 import json
+import logging
+import os
+import sys
+import tempfile
+import threading
 import time
-from typing import Any, Dict, Optional
+from collections import OrderedDict
 from dataclasses import dataclass
+from datetime import date, datetime
+from enum import Enum
+from typing import Any, Dict, Generic, List, Optional, Tuple, Type, TypeVar
+
 from filelock import FileLock
 
 logging.getLogger("filelock").setLevel(logging.INFO)
-
-from typing import TypeVar, Generic, Type
-from typing import List
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


aix-ahmet:
## Code review findings (2026-09-09, head `515a133a`)

The four ticket defects are closed and the body's runtime claims held up under re-verification. The findings below are mostly regressions the hardening introduces, and two go to the PR's own performance claims. Each was reproduced against the PR head, not inferred. Correctness first.

|---|---|---|---|
| 1 | `aixplain/utils/asset_cache.py:148` | High | `atomic_write_private` calls `os.fchmod`, which does not exist on Windows before Python 3.13 (`requires-python >= 3.9`). The `AttributeError` is not an `OSError`, so it escapes the handler, the temp file is deleted, and `save()` logs and returns. Every cache write silently fails on Windows; `cache_utils.save_to_cache` the same way. The base branch's plain `open()`/`json.dump` worked there. CI is ubuntu-only and the new tests skip mode assertions on `nt`, so nothing catches it. Guard with `hasattr(os, "fchmod")` and fall back to `os.chmod` on the path. |
| 2 | `aixplain/utils/asset_cache.py:559` | High | `copy.deepcopy` raises `RecursionError` for any `Model` with `model_params` (`BaseParameters.__getattr__` recurses on the attribute-less instance `copy._reconstruct` creates), so `get()` silently falls back to a shallow copy that aliases `model_params`/`input_params`/`additional_info` with the shared entry. Reproduced: after `cache.add(m)`, `cache.get('x').model_params is m.model_params` is `True`. Any caller setting `model.model_params.temperature` on a "copy" rewrites the process-wide entry, and each hit pays a ~1000-frame recursion first. The isolation test uses a `Model` without `model_params`, so it passes. |
| 3 | `model_getter.py:91` | High | On a cache **miss**, `get()` stores the fetched `Model` in the shared cache and returns that same object; only the hit path copies. Two agents in one process with the same `llmId`: `build_llm(A)` misses, gets the shared entry, then does `setattr(llm, 'temperature', ...)` and assigns `model_params` on it; `build_llm(B)` hits, copies the *mutated* object, and B silently runs with A's temperature and params. Base used a per-call `AssetCache`, so the alias was discarded; `shared()` makes it live for the process. The `add_list` path at 84-87 has the same alias and also skips the `api_key` stamp. |
| 4 | `aixplain/utils/asset_cache.py:410` | High | `load()` now does the expiry comparison and `raw_data` iteration outside any `try`, so a malformed cache file raises out of `AssetCache.__init__`/`shared()`. Reproduced: `{"expiry": null, "data": {}}` → `TypeError`; `"data": []` → `AttributeError`. `agent_factory/utils.py:176` calls `AssetCache.shared(Model)` before its `try`, so every `AgentFactory.get` with tools fails until the user deletes the file. Base wrapped everything and self-healed. |
| 5 | `aixplain/utils/asset_cache.py:787` | High | `_is_redacted` runs on every dict key at every depth, but `model_params`/`input_params`/`output_params` are keyed by parameter **name**, so any parameter literally named `url`, `token`, `password`, `secret`, `authorization` or `backend_url` is silently dropped from the persisted entry. Reproduced: `input_params={'text':..., 'url':...}` serialises with only `text`. A scrape/fetch utility or connection tool with a `url` parameter is cached without it; next process, `ModelTool.validate_parameters` raises, `build_tool_safe` swallows it, and the tool vanishes for the 24h TTL. Redact by path (the known credential attributes on the asset), not by key name at any depth. |
| 6 | `agent_factory/utils.py:180` | Medium | Cache-key mismatch for slug ids: `get()` looks up `model_id.replace('/', '%2F')` while prefetch and `add` store under the raw `model.id`. For an agent with `llmId` `aixplain/openai/gpt-4o-mini/openai` and slug tool ids (the common real payload), prefetch warms keys the getter never reads, each pool worker misses, fetches, and saves under the `FileLock`, and the N+1 plus lock contention the PR says it removes are still there. On the next load `asset_id in cache` is true for the raw id, so prefetch skips them and nothing is ever warmed. The `get()` half is pre-existing; the prefetch half is new. |
| 7 | `model_getter.py:84` | Medium | The retained cold-cache branch `cls.list()` → `add_list()` → `save(merge=False)` now runs on a process-wide instance that never re-reads disk after `__init__`. `_expire_if_needed` clears memory only (its docstring says the next miss repopulates from the file, but nothing outside `__init__` calls `load()`), so a long-lived worker whose store expired replaces the on-disk file with one page of 20 arbitrary models and discards what other processes just wrote. If prefetch fails, N pool workers each see an empty store and each rewrite the file. `add_list` also keeps a stale `store.expiry` on an expired store, writing an already-expired file. |
| 8 | `model_getter.py:97` | Medium | The fallback after the cache branch raises no longer calls `cache.add(model)`, so with `use_cache=True` and a failing `cls.list()` (paginate 403/500) the store stays empty and every later lookup re-issues the failing paginate POST, with retries, before the direct GET. A 10-tool agent issues 10 failing paginates plus 10 GETs on every build while paginate is unhealthy; base cached the direct fetch so the second call hit. |
| 9 | `aixplain/utils/asset_cache.py:123` | Medium | `ensure_cache_folder` unconditionally `chmod`s the directory named by `AIXPLAIN_CACHE_FOLDER` to `0700` on every save, and `default_cache_folder` returns the override verbatim with no dedicated subdirectory. A user who sets it to `$HOME`, `.`, or a shared volume read by another uid gets that directory's group/other bits silently stripped, logged only at debug, and re-applied on every save so a manual fix is reverted. Apply the mode only to a directory the SDK created, or append an `aixplain/` subdirectory to the override. |
| 10 | `tests/functional/model/run_model_test.py:311` | Low | The two new `assert not Path('.cache').exists()` checks assert pre-existing CWD state, not the PR's behaviour: any checkout where the old SDK ran once still has a gitignored `./.cache/`, and the fixture only invalidates the new location, so both tests fail on such a machine even though the new code wrote nothing there. Related: nothing removes the legacy `./.cache/model.json` on existing installs, so the already-leaked key stays on disk after upgrade. Consider a one-time cleanup of the legacy file, and make the tests assert on `model.json` in the CWD rather than the directory. |

Refuted: the LLM-to-plain-`Model` downgrade on read-back is pre-existing and the body acknowledges it; "prefetch serves private models across API keys" cannot occur because `build_agent` never receives the caller's key.

Cleanups that survived verification but are not blocking: a duplicate redaction key set alongside `v2/client.py::_REDACT_KEYS`; `default_cache_folder` duplicating `default_experiment_cache_dir`; the `_serialized` shadow dict; `cache_utils.py` retained with zero importers; a full-file rewrite per `add()`.

PR #1090, opened 2026-09-08
fix(generator): write to aixplain/v1 paths and escape backend values (ENG-3435)

aix-ahmet:
## Summary

`generate.py` wrote to `aixplain/enums/generated_enums.py` and `aixplain/modules/pipeline/pipeline.py` — two directories the v1 reorg deleted. The 11,606 lines it is supposed to own (`aixplain/v1/enums/generated_enums.py`, `aixplain/v1/modules/pipeline/pipeline.py`) were therefore hand-maintained files stamped "PLEASE DO NOT EDIT" with no working generator behind them, and `generate.py` was referenced by nothing in `.github/`, `.pre-commit-config.yaml`, `pyproject.toml`, `AGENTS.md`, `README.md` or `pydoc-markdown.yml`.

**Worth calling out: the crash is the good failure mode.** On a clean checkout the write raises `FileNotFoundError`, as the ticket says. On a developer machine still carrying `aixplain/enums/__pycache__` residue from before the reorg the directory *does* exist, `open(..., "w")` **succeeds**, and the generator writes 11k lines into a gitignored path that never shows up in `git status`. The run reports success and produces nothing tracked. That silent success is harder to notice than the crash.

Fixing the path alone would have re-armed the latent injection hole, so the escaping lands with it.

### What changed

The generator is now split into a networked **fetch** step and a deterministic, offline **render** step:

| Invocation | Network | Credential | Writes |
| --- | --- | --- | --- |
| `python generate.py` (default `render`) | no | no | the three generated modules |
| `python generate.py render --check` | no | no | nothing; non-zero exit on drift |
| `python generate.py fetch` | **yes** | **yes** | `tools/generator/fixtures/*.json` |

A CI job that ran the generator as-is would need a live backend and a secret in ordinary CI — exactly what BUG-947 moved the functional suite away from, and a drift check whose input moves under it is a flake generator. Rendering from a committed fixture makes the `generator-drift` job credential-free and fully reproducible.

## Per-item status

| # | Ticket item | Status | Where |
| --- | --- | --- | --- |
| 1 | Update the path constants, add `os.makedirs` | Done — and resolved against the repo root, not the CWD, so running the script from any directory writes to the right place | `generate.py:42-50`, `write_modules()` |
| 2 | Templates emit `aixplain.v1.*` | Done — `aixplain.v1.base.parameters`, `aixplain.v1.enums`, `aixplain.v1.modules`; `enums_include.py` too | both templates |
| 3 | Emit via `repr()`, validate identifiers | Done — `py_literal`, `py_identifier`, `py_docstring`; `escape_quotes` and `none_to_none` deleted; every interpolation routed through a filter, including the previously bare, unquoted `{{ supplier.id }}` | `generate.py` |
| 4 | CI job fails on a non-empty `git diff` | Done — the job runs `python generate.py render --check`, which names the drifted files and touches nothing tracked | `.github/workflows/generator-drift.yaml` |
| 5 | One module per function family with a re-exporting `__init__.py` | **Out of scope — see below** | — |
| — | `aixplain/v2/enums_include.py` claimed to be generated but nothing generated it | Done — now genuinely generated from one declared name list, so the import block and `__all__` cannot disagree | `ENUMS_INCLUDE_TEMPLATE` |
| — | `TEAM_API_KEY` only; legacy `api.aixplain.com` default | Done — accepts both, `TEAM_API_KEY` first (the same precedence as `aixplain.v2.core.Aixplain`), and refuses to run when both are set but differ; default is now the repo's production URL | `get_config()` |

### Acceptance criteria

- [x] `python generate.py` completes and writes to `aixplain/v1/enums/generated_enums.py` and `aixplain/v1/modules/pipeline/pipeline.py`.
- [x] A CI job runs the generator and fails if the committed modules differ from a fresh render — with no `TEAM_API_KEY` in the job.
- [x] A supplier named `x"""; import os; os.system("...") #` produces valid, inert Python (asserted with `ast.parse` and by importing the rendered module in a subprocess and checking the payload's side effect never ran).
- [x] The generated files no longer import from `aixplain.{base,enums,modules}`.

## Regeneration diff

The load-bearing check is byte-level: after the fix, **`pipeline.py` differs from the committed file by exactly its two import lines and nothing else**. That is the proof the generator, not a human, owns these files again.

| Symbol | Before | After | Added | Removed |
| --- | --- | --- | --- | --- |
| `Function` members | 127 | 127 | — | **none** |
| `Supplier` members | 80 | 84 | `ALIBABA_CLOUD`, `ANTHROPIC`, `BYTEDANCE`, `OPENROUTER` | **none** |
| `Language` members | 284 | 284 | — | **none** |
| `License` members | 13 | 13 | — | **none** |
| `pipeline.py` node classes | 379 | 379 | — | **none** |
| `Pipeline` factory methods | 378 | 378 | — | **none** |

The only other data change is `Supplier.GOOGLE`'s display name (`"google"` → `"Google"`); its id is unchanged. **This is an intentional correction, not a regression**: a live probe shows every backend (dev, test, production) serves `"Google"`, so the committed enum was wrong, and the exact-dict `Supplier(...)` lookups in `aixplain/v1/modules/asset.py` and `aixplain/v1/modules/agent/tool/model_tool.py` that used to silently fall back on a Google-supplied asset now resolve to `Supplier.GOOGLE`. **Nothing is removed**, so no user holding a reference to a v1 enum member breaks. Had the fresh data dropped members, they would have been kept and listed as candidates for a separate deprecation pass.

### Fixture provenance — a deviation from the plan, deliberately

The plan said to capture the fixture from the dev backend. **Supplier ids are environment-specific**, and the committed `Supplier` members carry *production* ids:

| supplier | committed | production | dev | test |
| --- | --- | --- | --- | --- |
| `google` | 1769 | **1769** | 195 | 212 |
| `timechat` | 11688 | **11688** | 722 | — |

Committing a dev snapshot would silently rewrite every `Supplier.*.value["id"]` in the shipped SDK — a real behavioural change for users, dressed up as a data refresh. The fixture is therefore captured from production (`https://platform-api.aixplain.com`) with four read-only `GET`s. No credential is committed or written to any file. `fetch` **refuses a non-production `BACKEND_URL`** unless `--allow-nonprod` is passed, and writes `tools/generator/fixtures/provenance.json` (host, capture time, SHA-256 of each fixture); `test_committed_fixtures_are_provenanced_from_production` asserts the committed provenance names production, that the digests match the committed files, and that `google` carries its production id. A dev snapshot captured with `--allow-nonprod` still records the dev host, so committing it fails that test. The refresh workflow is in `tools/generator/fixtures/README.md`.

### Two things the render had to handle to reproduce the committed output

- **A `ruff format` pass.** The committed files are already `ruff format`-stable under the pinned `ruff 0.12.12` (verified: formatting a copy of each is a no-op), but the raw Jinja output is not — someone formatted these files after the last successful generation. Without the pass, "render reproduces the committed files" is unachievable and every drift check is noise. It also buys idempotency for free, since a formatted file is a fixed point. `ruff==0.12.12` is pinned in the `test` extra and in the `pytest-check` hook, and a test asserts all three pins agree.
- **A duplicated backend parameter.** The catalog serves `text` twice for `benchmark-scoring-asr` and `benchmark-scoring-mt` (the hypothesis, `name: "output"`, and the reference, `name: "reference"`, under one code). A node class can only expose one `text` attribute, so the generator has an explicit policy: a duplicate parameter code **fails the render** unless the function is listed in `KNOWN_DUPLICATE_PARAMETERS` *and* the copies agree on every emitted field, in which case the first is kept once, in **both** modules. That removes the second `text` entry from `FunctionInputOutput` for those two functions (the only regenerated-module change beyond the import lines); `Function.get_input_output_params()` keyed the list by code and already saw one entry, and `AssetNode._auto_populate_params` skips codes the node class already declares, so nothing downstream changes.

The `text-detection` function's `OutputType` is `"text "`, with a trailing space — surrounding whitespace is stripped before identifier validation (whitespace is never part of an identifier, so nothing is lost); anything else is rejected rather than rewritten.

### What the generator does and does not own

The generator produces the **v1** enums (`aixplain/v1/enums/generated_enums.py`), the **v1** pipeline node classes (`aixplain/v1/modules/pipeline/pipeline.py`) and the `aixplain/v2/enums_include.py` re-export shim, which nothing in `v2` imports. `v2`'s own `Function`/`Supplier`/`Language`/`License` enums are hand-maintained in `aixplain/v2/enums.py`, `aixplain/v2/__init__.py` imports from there, and `import aixplain` loads no v1 module (asserted by `tests/unit/test_v1_deprecation_warning.py`). A fixture refresh therefore changes what v1 exposes, not what `Aixplain().Model.search(suppliers=...)` accepts. Known gap, out of scope here: the v2 `Supplier` lists seven members and now lags the backend by the four suppliers this refresh added to v1. `AGENTS.md` and `AGENTS.ar.md` say this; an earlier revision of this PR said the opposite.

### Review follow-ups (second pass)

The generator code from the first pass had ten reproducible defects; all are fixed, each with a test:

1. `py_docstring` now wraps first and escapes each line, so a backslash straddling the wrap column can no longer be split into a line continuation plus a fresh escape (`SyntaxError`).
2. `populate_specs` assigns the stripped identifiers instead of discarding them, so a whitespace-padded function id renders `class TextGenerationInputs`, not `class TextGeneration Inputs`.
3. Duplicate parameter codes fail the render, with the documented `KNOWN_DUPLICATE_PARAMETERS` override (see above); both modules keep the same entry.
4. The pipeline template emits the **raw** backend code as `create_param(code=...)` and the stripped value only as the attribute name, so the node class and `FunctionInputOutput` agree on the wire-level key.
5. The render is staged and formatted in a scratch directory and only written into place once everything succeeded; a ruff failure leaves the tracked files untouched.
6. `fetch` loads `REPO_ROOT/.env` explicitly (no parent walk-up, the BUG-939 pattern), prints the backend host it captured from, and warns when that host is not production.
7. `api_request` sends with `timeout=(10, 30)` and routes `BACKEND_URL` through `aixplain.utils.url_safety.validate_config_url` (lazy import, fetch path only): a plain-`http` backend is refused before the key goes on the wire.
8. Non-finite floats are refused both at fixture load (`parse_constant`) and in `py_literal` (`math.isfinite`), instead of rendering a bare `nan` that raises `NameError` on import.
9. A null description renders as `""`, as the old generator did, so `ModelTool`/`build_tool_payload` keep sending `""` rather than `null`.
10. `get_config()` mirrors `core.py`'s precedence (`TEAM_API_KEY` first) and refuses when both variables are set and differ.

Also: the dead `populate_data_types` is gone; the over-120 lines are wrapped; a missing `supplier.id` / `license.id` / required function field fails with the field and record named (`suppliers.json[3] ('google') is missing 'id'`) rather than Jinja's `Undefined`.

### Review follow-ups (third pass)

Nine further findings, all fixed with tests; the regenerated modules are **byte-identical** (no committed value has padding, a control character or a trailing quote, and the duplicate-parameter policy keeps the same entry):

1. **Config resolved once, after the SDK import.** `import aixplain` runs `load_dotenv(<cwd>/.env)`, so `fetch` read the host before that import (for the banner) and `api_request` read it again after (for the request). `resolve_fetch_config` loads the repo-root `.env`, imports the SDK, then reads the environment once and passes `api_key`/`backend_url` into `api_request` explicitly. Tested in a subprocess with a real CWD `.env` and no `load_dotenv` stub: the printed host equals the requested host and the key comes from that `.env`.
2. **`DataType` validated.** `dataType`/`InputType`/`OutputType` are checked against a frozen copy of the `DataType` values (a test keeps the copy equal to the enum, without the render path importing `aixplain`); an unknown type fails the render naming the function. A new test executes the rendered `pipeline.py` against the real package in a subprocess and checks the resolved members.
3. **Descriptions.** A non-printable character (`\x00`, `\x1b`, U+200B) in a description fails the render with the function named; the trailing-quote fix-up that double-escaped a description ending in `"""` is removed (both template sites already end the docstring on a new line). Tests for both endings.
4. **Production gate and provenance**, see the provenance section above. The old warn-and-continue test is replaced by a refusal test, a `--allow-nonprod` test and the committed-provenance test. The provenance for the existing fixtures was derived from the fixture contents (production supplier ids) and the fixture file times, since `fetch` was not re-run.
5. **`enumify` strips before slugifying**, so `google ` and `google` are one member and the uniqueness check sees the collision. The padded-id test now asserts `Function.TEXT_GENERATION` on the enum side too.
6. **Explicit LF writes** for the staged render and the fixtures (`open(..., newline="\n")`; `Path.write_text(newline=)` is 3.10+). The test emulates Windows by making text-mode writes without `newline=` produce CRLF, then asserts no `\r` reaches the written files.
7. **Docs corrected**, see the section above.
8. **Duplicate-parameter policy tightened.** The allow-list is inputs-only (outputs get an empty one), fails on an entry the catalog no longer exercises (attribute or whole function), and pins the `name` of each copy: `{"text": ("output", "reference")}`. Adding `name` to the equality set would have made the two `benchmark-scoring` copies "differ", because they *are* two inputs served under one code; pinning the exact pair instead keeps the one entry both modules have always shipped while making a changed or vanished pair fail loudly.
9. **All-or-nothing fetch.** Every endpoint is fetched before any fixture is written; a failure on the third endpoint leaves all four untouched and writes no provenance (tested).
10. *(Low, not taken.)* `get_config`/`api_request` were kept rather than replaced by `AixplainClient`: the fix for 1 resolves the config in the generator and keeps `requests.get` as the stubbing seam for the transport tests, and with 9 in place a transient 5xx now costs a re-run rather than a half-written fixture set.

**Drift-job decision.** The dedicated `generator-drift` job is kept and now runs `python generate.py render --check` instead of `render` + `git diff --exit-code`: it is the clearer CI signal (a named context in `docs/ci/required-checks.md` that fails fast with the drifted file names, independent of the rest of the unit suite), and it no longer writes into the checkout. `tests/unit/test_generator.py` keeps its byte-equality assertion as the offline, local (pre-commit) form of the same guarantee.

## Out of scope

**Ticket item 5** — emitting one module per function family with a re-exporting `__init__.py` — is not in this PR. It is a large refactor of an unmaintained v1 file, it changes the public import surface for every v1 and v2 consumer of `Function`/`Supplier`, and it would swamp the security fix and the CI gate. It deserves its own ticket.

Also deliberately excluded: removing enum members the backend no longer returns (none here, but the policy matters for the next refresh), and any `pipeline.py` change beyond what the generator needs to reproduce it.

## Test evidence

`tests/unit/test_generator.py` — 67 tests, no network, no credentials:

```
$ python -m pytest tests/unit/test_generator.py -q
67 passed in 2.50s
```

Covering: byte-for-byte reproduction of all three generated modules; `render --check` green on a clean tree; idempotency; no legacy imports in the output; the injection payload renders to valid Python and is inert when imported in a subprocess (asserted by blanking every string constant via `ast` and checking no `import os` survives as *code*); unusable identifiers (`3d-scan`, `3.5-turbo`, `!!!`, `class`, `café`) fail the render with the offending value named; colliding member names fail naming both; the render makes no network call and runs with every credential env var unset; both API-key env vars are honoured; the ruff pins agree. Second pass: wrap-before-escape round-trips a backslash token straddling the wrap column through `ast`; padded function ids and parameter codes render stripped identifiers with raw wire codes on both sides; the known duplicate is kept once in both modules, an unlisted duplicate and a listed-but-differing one both fail naming the copies; a failed format leaves the target files untouched; a null description renders `""`; a missing `id` names the field and record; NaN is refused at load and in `py_literal`; `fetch` passes the timeout, refuses `http://`, loads only the repo-root `.env` and prints the host; `TEAM_API_KEY` precedence and the conflicting-keys refusal. Third pass: the CWD-`.env` subprocess test; `DataType` agreement, unknown types and the executed pipeline module; control characters and trailing quotes; the production refusal, `--allow-nonprod`, the CLI wiring and the committed provenance; padded enum members; CRLF emulation; the inputs-only, name-pinned, stale-entry duplicate policy; the all-or-nothing fetch.

Full suite (Python 3.9, via the `pytest-check` pre-commit hook):

```
$ pre-commit run --all-files
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check for merge conflicts................................................Passed
check for added large files..............................................Passed
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
pytest-check.............................................................Passed
```

Local `python -m pytest tests/unit` (Python 3.12, all extras installed): 2812 passed, 0 failed. `python generate.py render --check` is clean under the pinned ruff 0.12.12.

Also verified by hand: `python generate.py fetch` against production reproduces the committed fixtures byte-for-byte, so the fixtures are genuinely the fetch output and the stable-ordering merge works; `python generate.py` run from an unrelated working directory with `TEAM_API_KEY`, `AIXPLAIN_API_KEY` and `BACKEND_URL` unset writes to the repo, not the CWD.

## Jira

https://aixplain.atlassian.net/browse/ENG-3435


aix-ahmet:

The design is right: fixtures separate fetch from render, the drift job runs offline, escaping landed before the write path. The problems below are in the new generator code itself; each was reproduced against the PR head, not inferred. Correctness first.

One refutation worth recording in the PR body: the `Supplier.GOOGLE` display-name change `"google"` → `"Google"` is a **fix**, not a regression. A live probe shows every backend serves `"Google"`, so the old enum was wrong, and exact-dict `Supplier(...)` lookups in `asset.py:79` / `model_tool.py:195` that used to silently fall back now resolve correctly.

|---|---|---|---|
| 1 | `generate.py:207` | High | `py_docstring` escapes backslashes *before* wrapping with `break_long_words=True`, so a long token straddling column 79 can be split between the two characters of an escaped `\\`, leaving a backslash-newline continuation followed by a fresh escape sequence. Reproduced: `SyntaxError: (unicode error) truncated \UXXXXXXXX escape` in the rendered `pipeline.py`. Wrap the raw text first, then escape each line. |
| 2 | `generate.py:720` | High | `populate_specs` calls `py_identifier(function_name)` and `py_identifier(class_name)` for the side effect only and discards the stripped return, so a whitespace-padded function id passes validation but the raw padded string is interpolated: `class TextGeneration Inputs(Inputs):` is a `SyntaxError`. The PR itself documents the backend serving padded values. Assign the return value. |
| 3 | `generate.py:775` | High | `deduplicate_by_name` silently drops the second `text` parameter of `benchmark-scoring-asr`/`-mt`, but the fixture shows these are two distinct inputs (`name: "output"` vs `"reference"`) sharing one code. Worse, the enum side keeps the *last* duplicate while the pipeline side keeps the *first*, so the two generated modules disagree with no render failure. Fail on duplicate codes like `check_unique_members`, with an explicit override for the two known functions. |
| 4 | `generate.py:400` | High | The pipeline template emits the whitespace-*stripped* identifier as the wire-level `code=`, while the enums template emits the raw backend code. For a padded code the node class calls `create_param(code='text')` while `FunctionInputOutput` keys `'text '`; `AssetNode._auto_populate_params` then creates a second input. Use the raw code for `code=` and the stripped value only for the attribute name. |
| 5 | `generate.py:873` | High | `write_modules` overwrites all three tracked files and only *then* runs `ruff_format`, so a ruff failure (not installed — the repo's own `.venv` lacks it — or invalid Python from a bad value) leaves raw or broken output in place. Probe: 8,436 changed lines in `generated_enums.py`. Render and format into a temp dir, as check mode already does, and move into place only on success. |
| 6 | `generate.py:588` | Medium | `fetch()` calls bare `load_dotenv()`, which walks up parent directories — the exact pattern `aixplain/__init__.py` replaced for BUG-939. Nothing prints or checks which backend host the fixtures came from, though the README says supplier ids are environment-specific. A repo-root `.env` pointing at dev silently captures dev ids into the shipped enum and the drift job stays green. Use an explicit-path load and print (or assert) the host. |
| 7 | `generate.py:525` | Medium | `api_request` issues `requests.get(url, headers=headers)` with no timeout and sends the API key to whatever `BACKEND_URL` resolves to, bypassing `url_safety.validate_config_url` and `_request_with_retry` that gate every other SDK request. A stalled backend hangs `fetch` forever; a plain-`http` URL receives the key in clear. Add `timeout=(10, 30)` and a lazy import of `url_safety` in the fetch path. |
| 8 | `generate.py:176` | Medium | `py_literal` renders non-finite floats via `repr()` as bare `nan`/`inf`, which `json` accepts by default, so a `NaN` in a numeric field produces a module that passes ruff and the byte-diff gate but raises `NameError` at import. Contradicts the "fails the render instead" guarantee. Guard with `math.isfinite`, or `parse_constant` in `load_fixture`. |
| 9 | `generate.py:276` | Medium | A null description now renders as `None` where the old `escape_quotes(None)` rendered `""`. `ModelTool.__init__` copies it into `self.description`, and `build_tool_payload` emits `"description": null` in the agent-creation payload where `""` was sent before. Coerce `None` to `""` in the literal path. |
| 10 | `generate.py:495` | Low | `get_config()` prefers `AIXPLAIN_API_KEY` over `TEAM_API_KEY`, the opposite of `aixplain/v2/core.py:119`, and the generator imports nothing from `aixplain.*` so the conflict guard in `utils/config.py` never fires. With both set, fixtures are captured under a different account than every SDK call in the same shell. Mirror `core.py`'s order, or refuse when both are set and differ. |

Also worth addressing: `populate_data_types` is dead (the template never reads `data_types`); two new lines exceed the 120-char limit (`generate.py:346`, `test_generator.py:252`); a missing `supplier.id` surfaces as an opaque `Undefined value Undefined`; and the separate `generator-drift` workflow duplicates a gate the `pre-commit` job already runs via `test_generator.py`'s byte-equality assertion — either drop it or make it the single owner of that check rather than running both.

aix-ahmet:
## Code review findings, second pass (2026-09-08, head `0921b8eb`)

The ten first-pass findings are all closed. This pass reproduced a further set, concentrated in the fetch path and in what the render does *not* validate. One finding corrects a claim this PR itself introduced into AGENTS.md.

|---|---|---|---|
| 1 | `generate.py:559` | High | `fetch()` prints and validates the host from `get_config()` *before* `api_request()` lazily imports `aixplain`, whose package `__init__` runs `load_dotenv(CWD/.env)`. A CWD `.env` then silently changes the host and key actually used. Reproduced: banner says production with no warning, every request goes to dev with the dev key, fixtures get dev supplier ids. The tests stub both `load_dotenv` and `api_request`, so they cannot see it. Resolve config once, after the SDK import, and pass `api_key`/`backend_url` into `api_request`. |
| 2 | `generate.py:918` | High | `dataType`/`InputType`/`OutputType` are validated only as identifier-shaped strings, never as `DataType` members. A refresh with a new backend type renders, formats, passes `render --check` and the unit tests, and ships a `pipeline.py` that raises `AttributeError` at import. Because `aixplain/v1/modules/__init__.py` imports `.pipeline` eagerly, `import aixplain.v1.modules` breaks for every v1 user. No test executes the rendered pipeline module. A frozen set of `DataType` values checked in `populate_specs` closes it without importing `aixplain`. |
| 3 | `generate.py:231` | High | The function description is the one backend value embedded raw in a triple-quoted docstring rather than via `py_literal`. `py_docstring` escapes only backslashes and triple quotes, so control characters pass through: a description containing `\x00` is emitted, ruff exits 0, and `import aixplain` fails with "source code string cannot contain null bytes". Separately, the trailing-quote fix-up double-escapes a description ending in `"""`. Route through `repr`, or reject non-printable characters with the same attributed `GeneratorValueError`. |
| 4 | `generate.py:662` | High | `fetch` only *warns* on a non-production host and still writes all four fixtures, while `load_dotenv(REPO_ROOT/.env)` makes a developer's usual dev `.env` select dev. A bare `fetch && render` then regenerates every `Supplier.*.value['id']` with dev ids; `model_factory/utils.py:221` sends that id to `/sdk/models/paginate`, so supplier filtering silently breaks in the shipped SDK while the drift job stays green. The README provenance line is hand-maintained and unverified. Refuse non-production unless `--allow-nonprod`, and write a machine-readable provenance file a test asserts. |
| 5 | `generate.py:137` | Medium | `enumify` converts leading/trailing whitespace to underscores *before* `py_identifier` can strip it: `enumify('google ')` → `GOOGLE_`. A catalog containing both `google` and `google ` passes `validate_enum_data` and ships two members with the same wire code. Padded values demonstrably come from the backend (`functions[75].metaData.OutputType == 'text '`), and the PR's own padded-id test only checks the pipeline side. Strip before the whitespace substitution. |
| 6 | `generate.py:1047` | Medium | `render_formatted` stages with `Path.write_text` and no `newline=`, so on Windows the staged files get CRLF, ruff preserves it, and the byte compare is wrong: `render --check` reports every module drifted on a clean tree, and `render` rewrites the tracked files with an 11k-line CRLF diff. Use `open(..., newline='\n')` (`Path.write_text` gained `newline=` only in 3.10; `requires-python` is `>=3.9`). The fixture write at 679 has the same issue. |
| 7 | `AGENTS.md:79` (and `AGENTS.ar.md:53`) | Medium | The new "Code generation" section says the `Function`/`Supplier`/`Language`/`License` enums v2 uses live in the generated v1 module and that the generator is load-bearing for v2. It is not: `aixplain/v2/enums.py` defines its own hand-maintained copies (`Supplier` has 7 listed members, none of the 4 this PR adds), `aixplain/v2/__init__.py` imports from `.enums`, nothing imports `enums_include.py`, and a guard test asserts `import aixplain` loads zero v1 modules. A maintainer following this paragraph will expect a fixture refresh to change what `Aixplain().Model.search(suppliers=...)` accepts. Correct the text; note that v2's `Supplier` is now four members behind the backend as a separate gap. |
| 8 | `generate.py:863` | Medium | `KNOWN_DUPLICATE_PARAMETERS` is applied to the *outputs* pass as well as inputs, so a duplicate output code is silently collapsed instead of failing; an allow-list entry that never fires is not reported, so the special case can go stale unnoticed; and `PARAMETER_FIELDS` omits `name`, the one field (`output` vs `reference`) that shows the two copies are different inputs. Pass an empty allow-list for outputs, fail on an unexercised entry, and include `name` in the compared fields. |
| 9 | `generate.py:669` | Medium | `fetch` writes each fixture as its endpoint returns, so a 403 or timeout on a later endpoint leaves the set half-refreshed: `functions.json` and `suppliers.json` from the new catalog, the other two on the old snapshot. `fetch` exits non-zero but a subsequent `render` happily mixes two catalog states. Fetch all four first, then write, matching `render_formatted`'s all-or-nothing staging. |
| 10 | `generate.py:531` | Low | `get_config` and `api_request` re-implement key precedence, the default URL, `urljoin`, headers, retries and error handling that `aixplain.utils.config` and `AixplainClient` already provide, and the fetch path already imports the SDK for `validate_config_url`. The copy gets no 429/5xx retry, so a transient 503 fails `fetch` outright. Importing either module was verified not to load the generated modules. |

Refuted this pass: the `Supplier.GOOGLE` rename (a fix, all name lookups are `.lower()`); the stripped-identifier vs raw-`code` split (no padded codes exist and the split is what prevents a duplicate input); the dedicated drift workflow as duplication (it has a stated rationale as a separately requireable branch-protection context).

Lower-priority cleanups that survived but are not blocking: dead `is_list`/`default`/`is_fixed`/`is_reconstructor` spec fields (912); member names derived twice (926); languages keyed by `label` rather than `value` in `stable_order` (58); `raise e` instead of bare `raise` (578); missing Apache header on the new test file; `AGENTS.ar.md` missing the English section's closing paragraph.

PR #1091, opened 2026-09-09
ENG-3699: add a builtin_tools capability field to the v2 Agent

aix-ahmet:
## Summary

Adds `builtin_tools` to the v2 `Agent` — a plain list field that toggles the agent worker's built-in toolkits (`file`, `python`, `bash`) — plus `warnings` on the run result so a gated toolkit is observable from Python.

Jira: https://aixplain.atlassian.net/browse/ENG-3699 (parent story PROD-2907). Backend https://aixplain.atlassian.net/browse/ENG-3698 is now merged and deployed to dev. Read-after-write lag tracked separately as https://aixplain.atlassian.net/browse/ENG-3711.

```python
from aixplain import Aixplain

aix = Aixplain()
agent = aix.Agent(
    name="Workspace agent",
    instructions="Read the uploaded files and compute the summary statistics.",
    builtin_tools=["file", "python"],
)
agent.save()

result = agent.run("Summarize data.csv")
print(result.warnings)   # e.g. ["Built-in toolkit 'bash' is not enabled on this worker — skipped"]
```

## History — this branch was rewritten once

An earlier revision modelled toolkits as **agent asset rows** via a `BuiltinTool` class (`ToolableMixin`, a `toolkit` discriminator, `include`, per-toolkit settings). The platform wire contract changed: toolkits are a **capability toggle on the agent**, so there is nothing for a tool object to model. That class and all its plumbing were deleted — no `mixins.py` widening, no `_hydrate_tools` / `_normalize_tool_*` / `to_dict` special-casing, no `builtin_tool.py`. That redesign also removed the root cause of all six findings from the first review round; they are listed as obsolete below.

## Design

- **`builtin_tools: list[str] | None`** on `Agent`, serialized as `builtinTools`, hydrated from the same key in `get()`, round-tripping through `save()`.
- **`exclude=lambda v: v is None`** — an agent that never touched the field sends no `builtinTools`, so existing payloads are byte-identical and a fetched agent cannot null out server state it does not know about (the BUG-1093 lesson). Backend treats absent, `null` and `[]` alike as none.
- **`BuiltinToolkit` str-enum** backs the field for autocompletion; plain strings are equally valid. **No client-side validation of values** — an unknown or gated-off toolkit is dropped by the worker with a warning, never an error, so the SDK must not pre-empt that.
- **No `include`, no per-toolkit settings.** Python's workspace access is inferred by the worker: on exactly when both `file` and `python` are enabled.
- **`warnings: List[str]`** on `AgentResponseData`, decoded exactly like `artifacts` on the same dataclass, plus an `AgentRunResult.warnings` property.
- **`_step_unit_names`** lives in `aixplain/v2/agent.py` so both the unit and functional suites import it.

### This round — end-to-end dev run against the deployed backend

| # | Finding | Status |
|---|---|---|
| 1 | The functional test read `step["tool"\|"toolName"\|"name"\|"action"]`, but the engine names the executing unit at `step["unit"]["name"]`. On dev the parser returned an empty set, so `assert used & BUILTIN_STEP_TOOLS` failed even though `write_file` and `run_python` had run. | **Fixed** — `_step_unit_names` reads `unit.name` first, keeps the flat keys as fallback, and is now an importable helper in `agent.py` with unit coverage for the real dev step shape, precedence, and malformed input. |
| 2 | The engine puts `warnings: list[str]` on the run result, and the README promised "dropped at run time with a warning on the run result" — but `AgentResponseData` had no `warnings` field, so the warning was unobservable from Python. | **Fixed** — `warnings` added, mirroring `artifacts` decoding: `Optional` to silence the dataclasses-json `RuntimeWarning`, re-coerced in `__post_init__` for direct construction, non-string entries dropped, never raising on absence or a non-list. |
| 3 | `Agent.get()` immediately after `save()` trails the write by ~0.5–1 s on dev (7 of 44 creates measured), which would make the functional test flaky. | **Fixed** — a poll helper in the *functional test only* retries for ~8 s. No polling was added to the SDK, per the constraint. |
| 4 | **Found in review:** warnings were readable in only one of the two places the engine could put them. The poll path never passes the response through — `resource.py`'s `filtered_response` rebuilds it from a fixed allow-list of top-level keys, so a `warnings` emitted *beside* `data` rather than *inside* it is dropped before decoding. Confirmed against the real `poll()` with a mocked client: top-level placement yielded an empty `data.warnings`. Same failure mode as finding #1 — SDK looking in one place, engine using another — and it would have failed the gated-bash assertion on dev for a reason unrelated to gating. | **Fixed** — `AgentRunResult.warnings` prefers decoded `data.warnings`, accepts a hand-built dict `data`, and falls back to the raw body's top-level key, mirroring the existing `artifacts` property. |
| 5 | **Found in review:** the read-after-write poll helper retried a *mismatched* field but not a *failed* `get()`. The same lag can 404 a just-created agent, which escaped on the first attempt and defeated the helper's purpose. | **Fixed** — retries on exception, re-raises once the budget is spent. |
| 5b | **Found in final review:** `AgentRunResult.warnings`' top-level fallback called `.get` on `_raw_data` after an `or {}` guard, so a truthy non-dict body (plain-text response, a list) raised `AttributeError` inside the property — which `Result.__getattr__` swallowed and re-raised as a misleading "no attribute 'warnings'". | **Fixed** — the top-level key is read only from a dict; unit test added. |
| 6 | ~4 of 44 creates lost `builtinTools` entirely. | **Skipped, by instruction** — backend defect tracked as ENG-3711, not worked around here. |
| 7 | `tests/unit/v2/test_agent_response_data.py` has no test docstrings while sibling v2 test modules use them throughout (16 × `D103`). | **Skipped** — nothing gates it: pre-commit runs ruff only over `^aixplain/v2/` and CI has no ruff step. The test names are self-documenting; 16 docstrings restating them would be noise. Disclosed rather than silently changed. |
| 8 | README claim that `bash` "is accepted by the API but is currently disabled in every environment". | **Verified** — matches the design spec's *"`bash` stays accepted but gated … It stays disabled in every environment."* |

Re-verified after the refactor rather than assumed: the `_is_auto_deserialize_only` guard test still holds; moving normalization into `__setattr__` did not break no-clobber behavior; `agent.builtin_tools = "python"`, `save(builtin_tools="file")`, `from_dict` with a scalar, `dataclasses.replace`, and a save against a non-echoing backend all land on the right list with `is_modified` staying `False`.

### Earlier round — `BuiltinTool` design, now deleted

| # | Finding | Status |
|---|---|---|
| 9 | Non-string `toolkit` raised `TypeError: unhashable type` instead of the documented `ValueError`. | **Obsolete** — class deleted. |
| 10 | `include` / `deny_patterns` aliased the caller's list, so post-construction mutation bypassed validation. | **Obsolete** — no `include` in the new contract. |
| 11 | `as_tool()` spliced `_extra` values in by reference, sharing a list between the tool and every emitted payload row. | **Obsolete** — no `_extra` replay. |
| 12 | `to_dict()` lost the `type` discriminator to dataclasses-json. | **Obsolete** — no discriminator. |
| 13 | The `ToolableMixin` contortion, the `_validate_dependencies` "has `id` but falsy" trap, `_save_subcomponents` calling `save()`, and `_hydrate_tools` ordering around the `context is None` early return. | **Obsolete** — a plain agent field is not walked by any of those paths. |
| 14 | No client-side bounds checks on `timeout_s` / `max_read_bytes`. | **Obsolete** — no per-toolkit settings exist. |
| 15 | Adding a dataclasses-json `decoder` to `builtin_tools` would make every `save()` null the list out (`exclude` + `decoder` trips `resource.py`'s `_is_auto_deserialize_only`, and the backend does not echo `builtinTools`). | **Fixed and pinned** — decoder deliberately not registered; a metadata test guards against the plausible future "tidy up response parsing" change. |

## Testing

- `tests/unit/v2/test_agent_builtin_tools.py` — `to_dict()` emits `builtinTools` and omits it when unset; `get()` hydration; full `save()` / `get()` round-trip; a builtin-only agent saves without tripping `_validate_dependencies` or `_save_subcomponents`; unknown strings pass through untouched; scalar/tuple/set shapes; no aliasing of the caller's list; a create response without the key leaves the field intact; the field-metadata pin.
- `tests/unit/v2/test_agent_response_data.py` — `warnings` decoded from the payload, absent/`null`/non-list/non-string entries; direct construction; both wire placements through `AgentRunResult.warnings`; `_step_unit_names` against the real dev step shape, flat-key fallback, `unit.name` precedence, and malformed input.
- `tests/functional/v2/test_agent_builtin_tools.py` — 3 tests incl. the gated-bash warning assertion. **Not run** — needs live credentials; `builtinTools` is dev-only while CI's functional legs target test/prod.

```
$ uv run --python .venv/bin/python pytest tests/unit/v2 -q
1822 passed in 10.54s

$ uv run --python .venv/bin/python pytest tests/functional/v2/test_agent_builtin_tools.py --collect-only
3 tests collected

$ uv run --python .venv/bin/python ruff check aixplain/v2/        All checks passed!
$ uv run --python .venv/bin/python ruff format --check aixplain/v2/   30 files already formatted
```

Wider `tests/unit` was green at review time (2863 passed). `uv.lock` is deliberately not part of this diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,10 @@ Available tools per toolkit: `file` — `read_file`, `list_directory`, `glob`,
 `ValueError` naming the settings that toolkit accepts. Anything you leave unset
 keeps the worker's default.
 
+`include` distinguishes three states: omit it for every tool in the toolkit,
+pass a subset to narrow it, or pass `[]` to attach the toolkit with no tools at
+all.
+
 > The `bash` toolkit must be enabled per deployment. Where it is not, the backend
 > rejects an agent that attaches it.
 
--- a/README.md
+++ b/README.md
@@ -83,6 +83,47 @@ print(result.data.output)
 
 > Runs return typed objects — read outputs with `result.data.output`, not dict indexing.
 
+### Attach built-in toolkits
+
+Agents can use the worker's built-in toolkits — a sandboxed filesystem, a Python
+interpreter and a shell — without onboarding any asset. Attach them with
+`BuiltinTool`, which validates the toolkit, the tool names and the settings before
+anything reaches the backend.
+
+```python
+from aixplain import Aixplain
+from aixplain.v2 import BuiltinTool
+
+aix = Aixplain()  # reads AIXPLAIN_API_KEY from the environment
+
+agent = aix.Agent(
+    name="Workspace agent",
+    instructions="Read the uploaded files and compute the summary statistics.",
+    tools=[
+        # Omit `include` to expose every tool in the toolkit.
+        BuiltinTool(toolkit="file", include=["read_file", "glob", "grep"]),
+        BuiltinTool(toolkit="python", timeout_s=30),
+    ],
+)
+agent.save()
+
+# Toolkits read back as objects, so a setting is one attribute away.
+fetched = aix.Agent.get(agent.id)
+fetched.tools[1].timeout_s = 60
+fetched.save()
+```
+
+Available tools per toolkit: `file` — `read_file`, `list_directory`, `glob`,
+`grep`, `write_file`, `edit_file`; `python` — `run_python`; `bash` —
+`run_command`. Settings are per toolkit too (`file`: `max_read_bytes`,
+`max_results`; `python`: `timeout_s`, `expose_files`; `bash`: `timeout_s`,
+`max_output_bytes`, `deny_patterns`), and using one on the wrong toolkit raises a
+`ValueError` naming the settings that toolkit accepts. Anything you leave unset
+keeps the worker's default.
+
+> The `bash` toolkit must be enabled per deployment. Where it is not, the backend
+> rejects an agent that attaches it.
+
 ### Build a multi-agent team
 
 ```python
--- a/aixplain/v2/__init__.py
+++ b/aixplain/v2/__init__.py
@@ -5,6 +5,7 @@ from .rlm import RLM, RLMResult
 from .utility import Utility
 from .agent import Agent, Artifact, Budget, ContextOverflowStrategy
 from .tool import Tool
+from .builtin_tool import BuiltinTool
 from .skill import Skill
 from .actions import Input, Inputs, Action, Actions
 from .integration import TriggerTypeSpec, TriggerEventOption, TriggerTypes
@@ -102,6 +103,7 @@ __all__ = [
     "Budget",
     "ContextOverflowStrategy",
     "Tool",
+    "BuiltinTool",
     "Skill",
     "Resource",
     "File",
--- a/aixplain/v2/agent.py
+++ b/aixplain/v2/agent.py
@@ -1,5 +1,6 @@
 """Agent module for aiXplain v2 SDK."""
 
+import copy
 import json
 import logging
 import re
@@ -18,7 +19,7 @@ from .model import Model
 from .file import File
 from .skill import Skill
 from .mixins import ToolableMixin
-from .builtin_tool import BuiltinTool
+from .builtin_tool import BUILTIN_TYPE, TOOLKIT_TOOLS, BuiltinTool
 from ..utils.user_info_utils import build_run_metadata
 
 from .resource import (
@@ -965,6 +966,12 @@ class Agent(
         if isinstance(tool, str):
             return {"id": tool}
         if isinstance(tool, dict):
+            if not tool.get("id"):
+                # An id-less row *is* its own identity: a builtin row that
+                # stayed a dict, or any future inline-config tool. Collapsing
+                # these onto {"id": None, ...} would make two different rows
+                # look identical and hide in-place edits from ``is_modified``.
+                return copy.deepcopy(tool)
             return {"id": tool.get("id"), "type": tool.get("type")}
         return {"id": getattr(tool, "id", None), "type": getattr(tool, "type", None)}
 
@@ -1769,38 +1776,42 @@ class Agent(
         normal ``Tool.get``). Asset hydration requires a client context; without
         one (e.g. an unbound ``Agent`` in unit tests) the raw entries are kept
         as-is. Built-in toolkit rows reference no asset, so they are converted
-        first and unconditionally — see :meth:`_hydrate_builtin_entry`.
+        whether or not a context is present.
         """
         if not self.tools:
             return
-        self.tools = [self._hydrate_builtin_entry(entry) for entry in self.tools]
         context = getattr(self, "context", None)
-        if context is None:
+        if context is None and not any(self._is_builtin_row(entry) for entry in self.tools):
+            # Nothing to convert: leave the caller's own list object in place.
             return
         self.tools = [self._hydrate_tool_entry(entry, context) for entry in self.tools]
 
     @staticmethod
-    def _hydrate_builtin_entry(entry: Any) -> Any:
-        """Convert a persisted ``{"type": "builtin", ...}`` row into a ``BuiltinTool``.
-
-        Keeps ``agent.tools`` symmetric with what the caller passed: a toolkit
-        attached as a :class:`~aixplain.v2.builtin_tool.BuiltinTool` reads back as
-        one after ``get()``. Needs no network and no client context. Best-effort,
-        like the asset hydration below: a row that cannot be mapped stays a dict
-        and still serializes through ``_normalize_tool_for_api``.
-        """
-        if not isinstance(entry, dict) or entry.get("type") != "builtin":
-            return entry
-        try:
-            return BuiltinTool.from_api_dict(entry)
-        except Exception:
-            # Never let hydration break construction — fall back to the raw dict.
-            return entry
+    def _is_builtin_row(entry: Any) -> bool:
+        """True for a raw ``{"type": "builtin", ...}`` row that still needs converting."""
+        return isinstance(entry, dict) and entry.get("type") == BUILTIN_TYPE
 
     def _hydrate_tool_entry(self, entry: Any, context: Any) -> Any:
-        """Hydrate one ``tools`` entry into a Tool/Model object (best-effort)."""
+        """Hydrate one ``tools`` entry into a BuiltinTool/Tool/Model (best-effort)."""
         if not isinstance(entry, dict):
             return entry  # already a Tool/Model/Integration object or a string id
+        if self._is_builtin_row(entry):
+            # Must return before the asset branch below: a builtin row that also
+            # carried an id would otherwise be rebuilt as a catalog Tool and
+            # saved back as ``{"type": "tool", "assetId": ...}``, losing the
+            # toolkit config and pointing at an asset that does not exist.
+            try:
+                return BuiltinTool.from_dict(entry)
+            except (ValueError, TypeError):
+                # An unknown toolkit may come from a backend newer than this
+                # SDK, so the row is kept and replayed verbatim. A row whose
+                # toolkit this version *does* know is malformed, and staying
+                # silent would ship it to the backend.
+                if isinstance(entry.get("toolkit"), str) and entry["toolkit"] in TOOLKIT_TOOLS:
+                    raise
+                return entry
+        if context is None:
+            return entry
         tool_id = entry.get("id")
         if not tool_id:
             return entry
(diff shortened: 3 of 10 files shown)


aix-ahmet:
## Review: ENG-3699 `BuiltinTool`

Every item below was reproduced against `4ae33e0` with a plain `Aixplain(api_key="x", backend_url="https://example.test")` client. Ordered by severity.

### Bugs

**1. A string `deny_patterns` is exploded into single-character regexes.** [`builtin_tool.py#L161`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/builtin_tool.py#L161) snapshots with `list(...)` before any type check, and `from_api_dict` does the same on fetch.
```python
BuiltinTool(toolkit="bash", deny_patterns="rm -rf").as_tool()["deny_patterns"]
# ['r', 'm', ' ', '-', 'r', 'f']
```
Saved as-is, the worker compiles six one-character regexes and blocks almost every command. A persisted string row goes through the same path on `get()`, so an unrelated `save()` persists the corruption. Reject non-list values (a bare string is the obvious mistake) instead of iterating them; same for `include` at [`#L134`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/builtin_tool.py#L134), which currently splits `include="read_file"` into characters and reports them as unknown tool names.

**2. `Agent.to_dict()` loses the `type` discriminator.** `BuiltinTool` is a dataclass, so dataclasses-json recurses into it from [`resource.py#L288`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/resource.py#L288) and emits the field set (`toolkit`, every `None` setting, `_extra`) with no `"type": "builtin"`.
```python
d = Agent(name="y", tools=[BuiltinTool(toolkit="python", timeout_s=5)]).to_dict()
d["tools"][0]          # {'toolkit': 'python', 'timeout_s': 5, '_extra': {}, ...}
Agent.from_dict(d).build_save_payload()["tools"][0].get("type")   # None
```
`from_dict` cannot rehydrate it (no `type`, so [`_hydrate_builtin_entry`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/agent.py#L1792) passes it through) and the typeless row is sent to the backend. Before this PR a dict-tools agent round-tripped through `to_dict`/`from_dict` losslessly. `to_dict` needs to emit `as_tool()` for these entries, the same way `_get_serializable_state` already does.

**3. `include=[]` is rejected, but it is a valid persisted value.** [`#L135`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/builtin_tool.py#L135) raises on an empty list. The worker treats an explicit `[]` as "zero tools from this toolkit" (`payload_adapter._map_tools`), so a saved row with `include: []` cannot hydrate: `get()` silently returns a plain dict, `fetched.tools[0].max_read_bytes = ...` raises `AttributeError`, and the user has no way to construct the equivalent object. Allow `[]` and let `None` keep meaning "all".

**4. The hydration fallback swallows validation and collapses identity.** [`agent.py#L1795-L1798`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/agent.py#L1795-L1798) catches every exception from `from_api_dict` and keeps the raw dict. Two consequences:
- A user-supplied dict with a bad toolkit or include name is sent to the backend unchanged, so the client-side validation the README promises only applies when the user happens to construct a `BuiltinTool`:
  ```python
  Agent(name="x", tools=[{"type": "builtin", "toolkit": "file", "include": ["run_python"]},
                         {"type": "builtin", "toolkit": "sql"}]).build_save_payload()["tools"]
  # both rows pass through verbatim, no error
  ```
- Every fallback dict reduces to `{"id": None, "type": "builtin"}` in [`_tool_identity`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/agent.py#L967), so editing `row["include"]` in place, or swapping one fallback toolkit for another, leaves `is_modified` False and `before_run()` skips the implicit save.

Narrow the except to `ValueError` on the fetch path only, and raise for dicts the user passed to the constructor. For identity, treat "no id" as "the whole normalized row" regardless of whether the entry is a dict or an object.

**5. `_extra` replays identity keys and can override validated fields.** Only `workspace_root` is in `_SERVER_ONLY_KEYS`, so [`from_api_dict`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/builtin_tool.py#L214-L218) keeps `id`, `assetId`, `name`, `status` and [`as_tool`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/builtin_tool.py#L173-L180) re-emits them, contradicting `test_a_builtin_row_carries_no_asset_identity`:
```python
BuiltinTool.from_api_dict({"type": "builtin", "toolkit": "file", "id": "abc", "assetId": "abc"}).as_tool()
# {'type': 'builtin', 'toolkit': 'file', 'id': 'abc', 'assetId': 'abc'}
```
Because `_extra` is a public init field ([`#L112`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/builtin_tool.py#L112)) and is applied before the structural keys, it also bypasses `__post_init__` entirely:
```python
BuiltinTool(toolkit="file", _extra={"type": "model", "toolkit": "bash"}).as_tool()
# {'type': 'model', 'toolkit': 'bash'}
```
Make `_extra` `init=False`, set it after construction in `from_api_dict`, apply it after the structural keys in `as_tool`, and strip identity keys on the way in. `compare=False` on the same field also makes `==` disagree with `as_tool()`: two tools that serialize differently compare equal.

**6. A builtin row that fails hydration and carries an id becomes an asset tool.** With a bound client, a row like `{"type": "builtin", "toolkit": "rust", "id": "abc"}` falls through the builtin fallback into [`_hydrate_tool_entry`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/agent.py#L1800), which resolves it as `context.Tool(id="abc")`, and the next save sends `{"type": "tool", "assetId": "abc"}`. The toolkit config is lost and the backend is asked to attach a non-existent asset. Return early for any `type == "builtin"` dict before the asset path.

### Quality

- **Naming.** [`from_api_dict`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/builtin_tool.py#L190) should be `from_dict` so `BuiltinTool` satisfies the existing `HasFromDict` protocol like `Inspector` does.
- **Wire type list.** Add `"builtin"` to `ToolDict.type` at [`mixins.py#L51`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/mixins.py#L51); it is the one declared list of tool discriminators and `as_tool` is typed to return it.
- **One hydration pass.** [`_hydrate_tools`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/agent.py#L1776-L1780) now walks `self.tools` twice and duplicates the try/except fallback. A `type == "builtin"` branch at the top of `_hydrate_tool_entry` (which already early-returns id-less dicts) removes `_hydrate_builtin_entry` and the second pass. It also restores the old guarantee that an unbound agent does not replace the caller's list.
- **Fetch-path strictness.** `__post_init__` validates `include` against the hardcoded `TOOLKIT_TOOLS` on fetch as well as on construction, so the first tool the worker adds makes every older SDK degrade fetched agents to dicts. Be strict where the user types the value and lenient (keep, warn) on `from_dict`.
- **Test duplication.** The `aix` fixture at [`test_agent_builtin_tools.py#L33`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/tests/unit/v2/test_agent_builtin_tools.py#L33) is the sixth verbatim copy in `tests/unit/v2`; and the contract rows at [`#L15`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/tests/unit/v2/test_agent_builtin_tools.py#L15) already diverge from the copy in `test_builtin_tool.py` (3 include names vs 6). One `conftest.py` for both.
- **Client attribute.** [`core.py#L66`](https://github.com/aixplain/aiXplain/blob/4ae33e03b7a18b72f1111233776164242cfa1116/aixplain/v2/core.py#L66) binds `BuiltinTool` as a bare class attribute while every sibling is rebound with a context in `init_resources`. Since it needs no client, exporting it only from `aixplain.v2` avoids the one-off.
- **Prose.** The module is roughly half docstrings and comments that restate the code; the repo convention is a one-line WHY at most.

tim-nelson:
I don't think we need a new type – these are tools, so let's use `aix.Tool`.

We already have local models for desktop (which we display with a "Local" tag), so let's do the same for tools. We can view the file, python and bash tools on web and create agents with them there, but those tools would only be available to an agent when run locally or via the desktop app.

```python
files = aix.Tool.get("aixplain/file")
python = aix.Tool.get("aixplain/python")

# Comes from the backend, not from a table in the SDK
list(files.actions)
# ['read_file', 'list_directory', 'glob', 'grep', 'write_file', 'edit_file']

# Cut down the same way as tavily in the README
files.allowed_actions = ["read_file", "glob", "grep", "write_file"]

# Settings are action inputs, so they are typed and you can list them
python.actions["run_python"].inputs["timeout"] = 30

agent = aix.Agent(
    name="Financial modelling agent",
    instructions="Read the attached accounts, build a three-statement model, write it to model.xlsx.",
    tools=[files, python],
)
agent.save()
```

PR #1136, opened 2026-09-17
Prod 2918 2917 2921 remove v1 and plain data config

elsheikhams99:
# Remove SDK v1; make v2 config take plain data

Closes PROD-2918, PROD-2917, PROD-2921.

**This is a breaking change and the first release without v1.** Version is bumped to `0.3.0`. The last release containing v1 is `0.2.48`, which stays on PyPI and stays installable.

## Why these three together

They all rewrite the same two files (`aixplain/__init__.py`, `aixplain/v2/__init__.py`) and they answer one question between them: *what does a caller have to import to use this SDK?* After this PR the answer is `from aixplain import Aixplain` and nothing else — no version segment, no second syntax, and no import at all for configuration, which is now plain dicts and strings.

Four commits, reviewable in order:

| Commit | Story | Files |
| --- | --- | --- |
| `73ad601` | PROD-2918 — remove v1 | 385 (−91k lines) |
| `309af7a` | PROD-2917 — API key limits as plain data | 8 |
| `1e439a6` | PROD-2921 — plain data across v2 | 20 |
| `cfdea94` | fixes from self-review | 7 |

## Two things that need a decision before merge

**1. Eight factories are removed with no replacement.** `MIGRATION.md` previously said the removal was *"contingent on closing these gaps first — we will not remove a capability that has nowhere to go."* This PR removes `IndexFactory`, `PipelineFactory`, `BenchmarkFactory`, `CorpusFactory`, `DataFactory`, `DatasetFactory`, `FinetuneFactory` and `WalletFactory` anyway, per PROD-2918. The guide now says so plainly and points affected users at `0.2.48`, but that promise is broken and those customers have nowhere to go. **Confirm this is intended before merging.**

**2. The meaning of `0` in API key limits is assumed, not confirmed.** PROD-2917 asks for this to be checked with the backend; I had no backend access. The implementation takes the reading that is safe either way — an unset dimension is *omitted* from the payload, so it cannot mean "blocked" or "unlimited" — and `APIKey._limits_to_api_dict` is the single place to change if the backend turns out to require all four fields on every write. **Worth confirming before release.**

## PROD-2918 — remove v1

Gone: `aixplain/v1/` (124 modules), the `_compat.py` redirects, the `aixplain` console script (a wrapper over the v1 model-onboarding factory), the `generate.py` generator and its fixtures, `aixplain/v2/enums_include.py`, and the eight `aixplain/utils` modules that existed only to serve v1. `url_safety`, `user_info_utils` and `config` stay.

`_compat.py` keeps its name and its meta-path finder but **redirects nothing**. It turns a legacy import into a `ModuleNotFoundError` that names the v2 replacement, the release to pin, and the guide:

```text
'aixplain.factories' was part of aiXplain SDK v1, which was removed in 0.3.0.
Construct a client and use its resources:
    from aixplain import Aixplain
    aix = Aixplain()
  AgentFactory       -> aix.Agent
  ModelFactory       -> aix.Model
  ...
  No v2 equivalent: IndexFactory, PipelineFactory, BenchmarkFactory, ...
The last release that shipped v1 is 0.2.48; pin it with "pip install 'aiXplain==0.2.48'".
Migration guide: https://github.com/aixplain/aiXplain/blob/main/MIGRATION.md
```

Both spellings are covered — `import aixplain.factories` through the finder, and `from aixplain import factories` through the package `__getattr__` (see the self-review section; they need different mechanisms).

CI: the six v1-only functional legs are deleted, leaving `v2`. The `generator-drift` workflow is deleted with the generator. `package-integrity` and the release workflow now assert that legacy paths *fail* in the built wheel, so a v1 leak becomes a red build.

## PROD-2917 — API key limits as plain data

```python
key = aix.APIKey.get("Production")     # a key ID, a key value, or a name
key.asset_limits = [{"model": "openai/gpt-5", "token_per_minute": 10_000, "token_type": "output"}]
key.save()
```

Typed as `APIKeyLimitsDict`, so a misspelled field is a type error; at runtime an unknown key raises and names the accepted fields, rather than being dropped (which used to mean "no limit"). `APIKeyLimits` and `TokenType` are unchanged and now import from the package root.

Two correctness fixes ride along, both worth a close read:

- **Unset is no longer zero.** Every dimension defaults to `None` and is omitted from the payload, so setting `token_per_minute` alone leaves the other three unrestricted. Previously all four were always sent, with `0` for the ones the caller never mentioned. *This changes the default:* `APIKeyLimits().token_per_minute` is now `None`, not `0`.
- **Key values are matched in full.** `get_by_access_key()` compared only the first and last four characters, so two keys sharing both resolved to whichever the backend listed first — and limits were written to a key the caller never named. If the backend returns *masked* keys and more than one is consistent with the value, it now raises instead of guessing.

## PROD-2921 — plain data across v2

Every enum and non-resource dataclass in `aixplain/v2` is classified as input, output or internal, and **the classification is recorded in [`docs/v2-plain-data.md`](docs/v2-plain-data.md)** so the next person does not repeat the audit: 31 enums (20 input, 4 output, 7 internal) and 47 non-resource dataclasses.

The rule: anything a caller constructs or passes in accepts plain data; anything that only ever comes back stays an object and is untouched.

```python
session = aix.Session(agent=agent, execution_config={"criteria": "be terse"})
agent.budget = {"max_cost": 0.5, "max_iterations": 10}
agent.context_overflow_strategy = "summarize"
trigger.configuration = {"type": "recurring", "repeat": {"every": 2, "unit": "hour"}}
```

Covers `ExecutionConfig`, `Budget`, `Task`, `TriggerConfiguration`, `TriggerRepeatRule`, `UtilityModelInput`. Shared coercion lives in the new `aixplain/v2/plain_data.py` and reads accepted names off the dataclass, so adding a field needs no second edit. Coercion runs in `__setattr__`, not only `__post_init__`, wherever a caller can assign after construction.

Input enums get a `Literal` alias (`PrivacyValue`, `SupplierValue`, …). All but `TokenType` already subclassed `str`, so the string form worked at runtime and only the annotation was missing.

**`aixplain/__init__.py` now re-exports `aixplain.v2.__all__`** via a star import rather than a copied list of ~130 names — it cannot drift, and IDEs resolve it through that same `__all__`. It costs no import time: importing `aixplain.v2.core` already executed the v2 package.

Two defects the audit turned up: `Agent(tasks=[Task(...)])` raised `AttributeError` (the constructor assumed every task was a dict), and a misspelled key in a `budget` or `execution_config` dict was dropped in silence.

Out of scope as specified: inspectors and evaluators, pending reimplementation.

## Self-review pass (`cfdea94`)

I reviewed the branch and found four defects in code it added. All are fixed and each now has a test the branch was missing.

- **`from aixplain import factories` lost all the guidance.** The meta-path finder covers `import aixplain.factories`, but CPython's `_handle_fromlist` probes with `hasattr` first and swallows a finder's `ModuleNotFoundError`, so this spelling reported a bare "cannot import name". Fixed in the package `__getattr__`. **Note for reviewers:** this deliberately makes `hasattr(aixplain, "enums")` *raise* rather than answer `False` — that is the mechanism, not a side effect, since `hasattr` only swallows `AttributeError`. A test pins it so it is not "fixed" into an `AttributeError` later.
- **`APIKey.get()` misreported unrelated failures.** Any by-ID failure became "No API key matches", so an expired credential or a 5xx sent the caller hunting for a key that was still there. The fallback still runs unconditionally — gating it on a status set would break get-by-name the day a backend answers 500 for a malformed id — but once nothing matches, a failure outside 400/404/422 is re-raised as itself.
- **`TriggerConfiguration.repeat`** coerced only in `__post_init__`, so `config.repeat = {...}` left a dict and `_hydrate_schedule_fields` later raised `AttributeError`.
- **`coerce_struct`** let a missing required field escape as `TypeError: Task.__init__() missing 1 required positional argument`.

The `APIKey` test stub also raised a bare `ResourceError` for an unknown id, which `AixplainClient` never does; it now raises `APIError(status_code=404)`, which is what made the `get()` status handling testable at all.

## Testing

- `pytest tests/unit` → **2263 passed, 3 skipped**.
- 3 pre-existing failures in `test_client.py::TestRateLimitAndJitteredRetry` (local urllib3 predates `Retry.retry_after_max`). Verified they fail on `development` too — unrelated to this PR.
- Built a wheel from a `.git`-less export, installed it into a clean venv: all 131 exports resolve, and `import aixplain.factories`, `from aixplain import factories` and `from aixplain import enums` all fail with the full guidance.
- `mypy` on a sample flags exactly the misspelled `TypedDict` keys and none of the correct ones.
- Ruff clean on `aixplain/` and the new tests.

New guard tests: `tests/unit/test_v1_removal.py` (replaces the two deprecation-era modules), `tests/unit/v2/test_api_key_plain_data.py`, `tests/unit/v2/test_plain_data_inputs.py`.

## Reviewing 405 files

Most of the diff is deletion. The code worth reading is small:

- `aixplain/_compat.py` — the whole removal-error mechanism
- `aixplain/__init__.py` — star re-export + the `__getattr__` guard
- `aixplain/v2/plain_data.py` — new, ~130 lines
- `aixplain/v2/api_key.py` — limits + `get()`
- `docs/v2-plain-data.md` — the recorded audit
- `.github/workflows/main.yaml`, `release.yaml` — the v1-leak assertions

Everything under `aixplain/v1/`, `docs/api-reference/python/aixplain/v1/`, `tools/`, and the deleted test files is bulk removal.

## Docs not in this repo

These pages named in the stories live in the GitBook docs repo and need a matching PR there:

- `docs/platform-overview/teams-and-governance/rate-limiting.md` (dict form for limits, string `token_type`)
- `docs/getting-started/quick-start.md:139` (Budget)
- `docs/assets/agents/agents.md:577, 823, 1044` (OutputFormat, ContextOverflowStrategy, ExecutionConfig)
- `docs/api-reference/swift/swift-api-reference/discoveressential.md:12` (still shows `from aixplain.factories import ModelFactory` above Swift code)

The migration-guide notebook (Drive `1y6maxqLFd8oQ5NY9fdlgRLGLt4SzOVOZ`), section 5, also needs updating — I had no Drive access.

Every in-repo equivalent is updated (`MIGRATION.md`, both READMEs, `AGENTS.md`, `docs/ci/required-checks.md`, the skill references), and a guard test fails if any repo doc reintroduces a removed import path.


aixplain/v2/agent.py
```diff
@@ -1119,6 +1164,17 @@ def _coerce_budget(cls, budget: Optional[Union[Dict, "Budget"]]) -> Optional["Bu
         if budget is None or isinstance(budget, Budget):
             return budget
         if isinstance(budget, dict):
+            # Accepted on either spelling, but an unknown key raises: a
+            # misspelled ``max_iteration`` used to pass through
+            # ``_normalize_budget`` untouched and then be dropped by the
+            # constructor below, i.e. silently mean "no cap".
+            unknown = sorted(
```
aix-ahmet: **Blocking — this unknown-key check also runs on the deserialization path, and breaks forward compatibility.**

`ExecutionConfig.budget` is annotated `Optional[Any]`, so `dataclasses_json` leaves the backend's `executionParams.budget` as a raw wire dict. `ExecutionConfig.__post_init__` then feeds that dict straight into `_coerce_budget` — which now raises on any key it doesn't recognise.

Verified against both branches:

```python
Session.from_dict({'id':'s','executionConfig':{'executionParams':{'budget':{'maxCost':1.0,'maxTokens':9}}}})

# development:  Budget(max_cost=1.0, max_duration_seconds=None, max_iterations=None)
# this branch:  ValidationError: Unknown budget field(s): maxTokens
```

The day the backend adds any field to the budget object, **every `Session.get()` and every session hydration in the SDK starts failing** — on data the SDK only reads. `_normalize_budget`'s own docstring still promises "Unknown keys are passed through unchanged (forward compatibility)", which is now false for this entry point.

The rejection is right and wanted — but only for caller-supplied input. It belongs on the input path (`Agent.__setattr__`, `ExecutionConfig.coerce`), not inside `_coerce_budget`, which both paths share. Every other struct in this PR avoids the problem because `dataclasses_json` decodes it before the coercion sees it; `budget` escapes only because it's typed `Any`.

aixplain/v2/session.py
```diff
@@ -486,12 +507,27 @@ def _fold_legacy_max_iterations(cls, kvs: Any) -> Any:
...
         if isinstance(value, cls):
             return value
         if isinstance(value, dict):
+            unknown = sorted(key for key in value if key not in _EXECUTION_CONFIG_KEYS)
+            if unknown:
+                raise ValidationError(
+                    f"Unknown execution_config field(s): {', '.join(unknown)}. "
+                    f"Accepted fields: {', '.join(_EXECUTION_CONFIG_FIELDS)}."
+                )
             return cls.from_dict(value)
```
aix-ahmet: **Blocking — `session.execution_config = {...}` raises on save.**

`coerce` is only reached from `__post_init__`, so a post-construction assignment leaves a raw dict and the save path dies:

```python
s = Session(agent="abc")
s.execution_config = {"criteria": "be terse"}
s.build_save_payload()
# AttributeError: 'dict' object has no attribute 'to_api_dict'
```

The constructor path is fine (`Session(agent="abc", execution_config={"criteria": "be terse"})` → `{'executionConfig': {'criteria': 'be terse'}}`). Unknown-key validation is skipped on the assignment path too — `s.execution_config = {"critera": "typo"}` is accepted silently, which is the exact failure mode this `coerce` was added to remove.

This is the same defect `cfdea94` fixed for `TriggerConfiguration.repeat`, and it violates the invariant `docs/v2-plain-data.md` records: *"Coercion happens in `__setattr__`, not only `__post_init__`, wherever a caller can assign the field after construction."*

Fix: a `Session.__setattr__` routing `execution_config` through `ExecutionConfig.coerce`, mirroring `Trigger.__setattr__`. Worth a test pinning the assignment path, since the existing coverage only exercises the constructor.

aixplain/v2/agent.py
```diff
@@ -713,7 +755,10 @@ class Agent(
 
     def __post_init__(self) -> None:
         """Initialize agent after dataclass creation."""
-        self.tasks = [Task.from_dict(task) for task in self.tasks]
+        # Dicts and ``Task`` objects both, on the user-facing field names. The
+        # previous ``Task.from_dict(task)`` assumed a dict and raised
+        # ``AttributeError`` on a ``Task`` the caller had constructed.
+        self.tasks = coerce_struct_list(self.tasks, Task, label="tasks", aliases=_TASK_WIRE_ALIASES)
```
aix-ahmet: **Blocking — `agent.tasks = [...]` silently sends the wrong wire keys.**

Same `__post_init__`-only gap as `Session.execution_config`, but the failure is quieter and worse: it doesn't raise, it sends a malformed payload.

```python
a.tasks = [{"name": "t", "instructions": "i", "expected_output": "e"}]
a.build_save_payload()["tasks"]
# [{'name': 't', 'instructions': 'i', 'expected_output': 'e'}]

Agent(..., tasks=[<same dict>]).build_save_payload()["tasks"]
# [{'name': 't', 'description': 'i', 'expectedOutput': 'e', 'dependencies': []}]
```

`tasks` is serialised through `dataclass_json`, which passes an unrecognised dict through untouched — so the user-facing names reach the backend instead of `description` / `expectedOutput`, and `dependencies` is dropped. `_TASK_WIRE_ALIASES` never runs.

Unknown keys are silently accepted here as well: `a.tasks = [{"nam": "t"}]` passes, while the constructor correctly raises `ValidationError: Unknown tasks field(s): nam`.

`docs/v2-plain-data.md` lists `agent.tasks[*]` as accepting `TaskDict`, so the doc currently overstates this. Fix by moving the `coerce_struct_list` call into `Agent.__setattr__` alongside the existing `budget` coercion.

aixplain/v2/api_key.py
```diff
@@ -333,29 +519,153 @@ def list(cls, **kwargs) -> List["APIKey"]:
...
-            ValidationError: If access_key is too short
-            ResourceError: If no matching key is found
+            ResourceError: If nothing matches, or if a name or masked key value
+                matches more than one key.
+            Exception: Whatever the by-ID fetch raised, if that failure was not a
+                verdict on the argument (an expired credential, a 5xx) and no
+                key matched the other two forms.
+        """
+        try:
+            return super().get(id, **kwargs)
```
aix-ahmet: **A live API key value ends up in the request URL path.**

The new contract is "a key ID, a full key value, or a key name", and this line tries the by-ID fetch *first* for all three. `GetResourceMixin.get` builds `path = f"{resource_path}/{encode_resource_id(id)}"` (`resource.py:1395`), so a caller following the documented usage —

```python
key = aix.APIKey.get("<their actual key value>")
```

— puts their secret into a URL path segment, where it reaches access logs, proxies, and error-tracking breadcrumbs. Before this PR, key-value lookup went only through `get_by_access_key()` → `list()` and never touched the URL.

The comment at line 565 reasons carefully about not echoing the argument in the *error message*, and `_mask()` exists for exactly that concern — but the request line itself is the bigger exposure and is unguarded.

Worth detecting a key-shaped argument (or simply trying the listing before the by-ID GET) so a secret never becomes a path segment. That also removes the wasted round-trip for the name and key-value forms.

aixplain/v2/agent.py
```diff
@@ -151,6 +153,40 @@ class ContextOverflowStrategy(str, Enum):
...
+    uncapped. Declared as a ``TypedDict`` so a plain dict still gets
+    autocomplete and a type error on a misspelled key, with nothing to import.
+    """
+
+    max_cost: float
+    max_duration_seconds: float
+    max_iterations: int
+
+
+class TaskDict(TypedDict, total=False):
```
aix-ahmet: **`total=False` here contradicts `Task`'s required fields, so mypy accepts a dict the runtime rejects.**

`Task` requires `name`, `instructions` and `expected_output` — none has a default — but `TaskDict.__required_keys__` is empty. A dict that type-checks clean therefore fails at runtime:

```python
Agent(name="n", description="d", tasks=[{"name": "t"}])
# ValidationError: Missing required tasks field(s): instructions, expected_output
```

The TypedDict is the type checker's contract for exactly this input, so it should be `total=True` with `NotRequired[...]` on `dependencies` only. Same issue in `code_utils.py:31` — `UtilityModelInput` requires `name` and `description`, `UtilityModelInputDict` marks both optional.

Contrast `BudgetDict`, `APIKeyLimitsDict`, `ExecutionConfigDict`, `TriggerConfigurationDict` and `TriggerRepeatRuleDict`, where `total=False` is correct because every field on the underlying dataclass defaults.

Note this also means `test_typed_dict_keys_are_all_optional` in `tests/unit/v2/test_plain_data_inputs.py` is asserting the wrong invariant — it should be "no key is required that the dataclass doesn't require", not "no key is required."

aixplain/__init__.py
```diff
@@ -103,4 +134,4 @@ def __dir__():
 # ``aixplain_v2`` is deliberately absent: it stays importable by name through
 # ``__getattr__`` above, but keeping it here would make ``from aixplain import *``
 # construct a client -- and therefore raise -- in a keyless environment.
-__all__ = ["Aixplain", "File"]
+__all__ = list(_V2_ALL)
```
aix-ahmet, line 137: **The star re-export shadows the builtin `TimeoutError`.**

`aixplain.v2.__all__` contains `"TimeoutError"`, and `aixplain.v2.exceptions.TimeoutError` derives from `AixplainV2Error`, **not** from `builtins.TimeoutError` — verified MRO is `TimeoutError → AixplainV2Error → Exception → BaseException`, and `issubclass(aixplain.TimeoutError, builtins.TimeoutError)` is `False`.

So after the star import this PR's ergonomics actively encourage:

```python
from aixplain import *
...
except TimeoutError:   # no longer catches socket/asyncio timeouts
```

`__all__` was `["Aixplain", "File"]` before, so this exposure is new. Cheapest fix is excluding `TimeoutError` from the package-level `__all__` (it stays importable by name); making it inherit from the builtin as well would be the more robust one.

Same shape, lower stakes, for the other very generic names now at the root: `Dataset`, `Metric`, `Input`, `Task`, `Experiment`, `Action`, `File`. Worth a deliberate decision about which of those belong in a star-importable `__all__`, given that "cannot drift" was the argument for `list(_V2_ALL)`.

aixplain/v2/agent.py
```diff
@@ -151,6 +153,40 @@ class ContextOverflowStrategy(str, Enum):
     SUMMARIZE = "summarize"
 
 
+#: The string form of :class:`OutputFormat`, accepted anywhere the enum is.
+OutputFormatValue = Literal["markdown", "text", "json"]
+
+#: The string form of :class:`ContextOverflowStrategy`, accepted anywhere the
+#: enum is -- ``agent.context_overflow_strategy = "summarize"``.
+ContextOverflowStrategyValue = Literal["truncate", "summarize"]
```
aix-ahmet: **18 of the 20 `Literal` aliases are attached to nothing.**

PROD-2921 asks that *"every input enum accepts its string values, typed as `Literal`"* and that *"a misspelled key is flagged by the type checker."* The aliases are all defined, exported, documented and drift-tested — but grepping `aixplain/` for annotation use of them finds only two live call sites: `APIKeyLimitsDict.token_type` and `UtilityModelInput.type`.

The other 18 — including the two defined here — are declared and never referenced. The fields `docs/v2-plain-data.md` names as their call sites still carry the old annotations:

- `agent.py:718` — `output_format: Optional[Union[str, OutputFormat]]`
- `agent.py:751` — `context_overflow_strategy: Optional[str]`

So neither typo is caught, at type-check time or runtime:

```python
agent.output_format = "markdon"               # accepted
agent.context_overflow_strategy = "sumarize"  # accepted
```

The runtime half of the story is delivered (these enums subclass `str`, so the string form always worked); it's the annotation half that's missing, and that was the stated point of adding the aliases. Cheapest fix is to annotate the fields — `Optional[Union[OutputFormatValue, OutputFormat]]` and `Optional[Union[ContextOverflowStrategyValue, ContextOverflowStrategy]]` — and likewise for `Privacy`, `Supplier`, `Function`, `Language`, `License`, `StorageType`, `SplittingOptions`, `IssueSeverity` and `ProgressFormat` at the sites the doc names.

If this is deliberately deferred, the doc and the story should say so rather than imply the aliases are wired up.

aixplain/v2/api_key.py
```diff
@@ -333,29 +519,153 @@ def list(cls, **kwargs) -> List["APIKey"]:
...
+        except Exception as by_id_error:
+            if not isinstance(id, str):
+                raise
+            # ``except ... as`` unbinds the name when the block ends, so keep a
+            # reference for the decision made after the fallback.
+            id_lookup_error = by_id_error
+            try:
+                candidates = cls.list()
```
aix-ahmet: **`**kwargs` is dropped on the fallback path.**

The docstring says *"`**kwargs`: Forwarded to the underlying request"*, and `super().get(id, **kwargs)` does forward them — but this `cls.list()` doesn't, so anything the caller passed is lost the moment the by-ID lookup misses. `get_by_access_key` gets this right (`cls.list(**kwargs)`).

Should be `cls.list(**kwargs)`.

aixplain/v2/api_key.py
```diff
@@ -333,29 +519,153 @@ def list(cls, **kwargs) -> List["APIKey"]:
...
+            except Exception:
+                # The listing is the fallback; if it fails too, the original
+                # per-ID failure is the more useful one to surface.
+                raise by_id_error from None
+
+        found = cls._resolve_by_key_value(id, candidates, required=False)
```
aix-ahmet: **Masked-value guessing runs before exact name matching.**

The ordering is: exact key value → *masked prefix/suffix guess* → name. A name is an exact equality comparison; a masked match is, by `_resolve_by_key_value`'s own docstring, *"a guess"*. Trying the guess first means a name argument can resolve through 4-char prefix/suffix matching before the exact comparison on line 567 ever runs.

There's also no length floor on this path. `get_by_access_key` enforces `len(access_key) >= 8` precisely so prefix and suffix can't overlap — `get()` calls `_resolve_by_key_value` directly and bypasses that guard, so a short name like `"key"` yields `prefix == suffix == "key"`.

Both are unlikely in practice, but the whole point of this commit was to stop resolving keys by prefix/suffix coincidence. Moving the `by_name` block above the `_resolve_by_key_value` call costs nothing and removes the class of problem.

MIGRATION.md
```diff
@@ -342,15 +348,15 @@ An internal abstract base class that other v1 factories inherit `get` from. It h
 
 ---
 
-## No v2 equivalent yet
+## No v2 equivalent
 
-These eight factories have **no v2 replacement today**. They remain fully supported in v1, and the removal date above is contingent on closing these gaps first — we will not remove a capability that has nowhere to go.
+These eight factories have **no v2 replacement**, and the removal took them with it. There is no v2 spelling to port them to.
```
aix-ahmet, line 353: **This is the line I'd want signed off before merge.**

The previous text committed to the opposite:

> These eight factories have **no v2 replacement today**. They remain fully supported in v1, and the removal date above is contingent on closing these gaps first — we will not remove a capability that has nowhere to go.

I confirmed the gap is total: `aixplain/v2/` contains no `Index`, `Pipeline`, `Wallet`, `Finetune`, `Benchmark`, `Corpus` or `Dataset` class. `IndexFactory` is the only Python path to knowledge bases / RAG — a headline capability, not a long-tail one — and "stay on `0.2.48`" means those users stop receiving SDK updates entirely.

The rewrite here is honest and points people somewhere real, which is the right way to break the promise if it's going to be broken. But PROD-2918 records no decision to break it. Worth a written sign-off on the story before this merges, rather than having it land as an implicit consequence of the removal.

aix-ahmet (commented):
## Review against PROD-2918 / PROD-2917 / PROD-2921

Two passes: the three stories' acceptance criteria run as executable checks, and a correctness pass over every non-deletion hunk.

**Verification.** `pytest tests/unit` on this head → **2266 passed, 3 skipped, 0 failed**. The three `TestRateLimitAndJitteredRetry` failures you reported don't reproduce here, so the suite is fully green. `0.2.48` is confirmed live on PyPI, so the pin guidance in `_compat.py`, `MIGRATION.md` and `CHANGELOG.md` is valid. Every behavioural claim below was run on this branch, and the regressions were run against `development` for comparison.

Most ACs pass exactly as written, and the shape of the work is right. Eight findings, three of which I'd fix before merge — plus the product decision you already flagged.

---

### Blocking

**1. `_coerce_budget`'s unknown-key check also runs on deserialization** (`agent.py:1171`) — the most serious one, and it isn't in the self-review list. `ExecutionConfig.budget` is typed `Any`, so the backend's raw wire dict reaches `_coerce_budget` and any unrecognised key now raises. `development` returns a `Budget`; this branch raises `ValidationError`. The day the backend adds a budget field, every `Session.get()` fails on data the SDK only reads.

**2. `session.execution_config = {...}` raises on save** (`session.py:531`)

**3. `agent.tasks = [{...}]` silently serialises the wrong wire keys** (`agent.py:761`) — no exception; the backend just drops the task's instructions and expected output.

2 and 3 are the same gap: coercion lives in `__post_init__` only, so the assignment path bypasses it. This is precisely the defect `cfdea94` fixed for `TriggerConfiguration.repeat`, and it violates the invariant `docs/v2-plain-data.md` records:

> Coercion happens in `__setattr__`, not only `__post_init__`, wherever a caller can assign the field after construction.

`Agent.budget`, `APIKey.global_limits`/`asset_limits`, `Trigger.configuration`, `TriggerConfiguration.repeat` and `UtilityModelInput.type` all got that treatment. `Session.execution_config` and `Agent.tasks` didn't — and both are listed in `docs/v2-plain-data.md` as dict-accepting inputs, so the doc currently overstates them.

### Worth fixing

**4. `APIKey.get()` puts a live key value in the request URL path** (`api_key.py:549`). The documented contract invites passing a key value, and the by-ID fetch is tried first, so the secret becomes a path segment — access logs, proxies, error-tracking breadcrumbs. `_mask()` carefully keeps it out of the *error message*; the request line is unguarded.

**5. `TaskDict` and `UtilityModelInputDict` are `total=False` but their dataclasses have required fields** (`agent.py:177`, `code_utils.py:31`). mypy accepts `{"name": "t"}`; the runtime rejects it.

**6. The star re-export shadows the builtin `TimeoutError`** (`__init__.py:137`). `aixplain.TimeoutError` does not subclass `builtins.TimeoutError` (verified MRO), so `from aixplain import *` silently breaks `except TimeoutError:` around socket/asyncio code.

### Non-blocking

**7. 18 of 20 `Literal` aliases are unattached exports** (`agent.py:161`) — the type-checker half of a PROD-2921 AC.

**8. `APIKey.get()`: `kwargs` dropped on the fallback, and masked-value guessing runs before exact name matching** (`api_key.py:557`, `:563`).

---

### What checks out

Verified by running each, not by reading:

- **PROD-2917 — all ACs pass.** Dict limits save; an unknown key raises naming the six accepted fields; `_limits_to_api_dict(APIKeyLimits(token_per_minute=10))` → `{'tpm': 10}`, so unset dimensions are genuinely omitted; `key.asset_limits[0].token_per_minute` returns `10` as an `APIKeyLimits` after a dict assignment; full key-value matching with an explicit raise on ambiguous masked keys; `from aixplain import APIKeyLimits, TokenType` works.
- **PROD-2921 — the dict surface is solid** on the construction path. `Agent(tasks=[Task(...)])` no longer raises; `{"max_iteration": 3}`, `{"critera": "x"}` and `{"nam": "t"}` all raise naming the accepted fields; missing required fields raise `ValidationError` rather than a bare `TypeError`; `trigger.configuration = {"type": "recurring", "repeat": {...}}` coerces both levels.
- **All 20 `Literal` aliases list enum *values*, not member names, and match their enums exactly** — checked against the enum definitions directly, not just via the drift test.
- **Hydration does not regress on unknown backend keys** for `Trigger` / `TriggerConfiguration` / `TriggerRepeatRule` / `APIKey` / `Agent.tasks` — `dataclasses_json` decodes those before the new coercion sees them. `ExecutionConfig.budget` is the one that escapes, which is finding 1.
- **PROD-2918.** Both `import aixplain.factories` and `from aixplain import factories` raise with the full guidance. No doc in the repo imports a v1 path. `namespaces = false` is safe — every remaining package directory carries an `__init__.py`. No residual references to the deleted `aixplain/utils/*` modules, `cli_groups`, `generate.py` or `enums_include.py`; the dropped `click`/`ruff` deps and the console script are unreferenced. The CI wheel assertions are internally consistent — every asserted substring is actually present in the corresponding `_PREFIX_GUIDANCE` message. `CHANGELOG.md` satisfies the release-notes AC.
- Dropping the `sys.modules` config sync from `core.py` is inert — nothing imports that module any more.
- `plain_data.py` reading accepted names off the dataclass, and the drift guards in `test_plain_data_inputs.py`, are the right call. `docs/v2-plain-data.md` is a genuinely good artifact and I'd keep the habit.

---

### On the two open questions

**The eight removed factories.** You're right to hold this. I confirmed the gap is total — `aixplain/v2/` has no `Index`, `Pipeline`, `Wallet`, `Finetune`, `Benchmark`, `Corpus` or `Dataset` class. `IndexFactory` is the only Python path to knowledge bases / RAG, a headline capability rather than a long-tail one, and "stay on `0.2.48`" means those users stop receiving SDK updates entirely. That's a product call, and PROD-2918 records no decision to make it. It should be signed off in writing on the story before this merges.

**The meaning of `0` in API key limits.** Genuinely unresolved and not settleable from the SDK side. The reading you chose is safe under either interpretation, and concentrating it in `_limits_to_api_dict` is the right shape. One thing to add to the release notes: this is also a *read*-path break — `APIKeyLimits().token_per_minute` is now `None`, so existing `limits.token_per_minute > 0` raises `TypeError`.

aix-ahmet (approved):

